PR #13, opened 2019-10-28
windows environment file

AyEnEss:
Hi Stefan.
this is the environment.yml file I was able to use on my windows machine and it worked. All I did was I deleted all the packages that were not able to resolve and then it worked without a hitch. check it out and if you think it would help the readers who use windows and are having troubles, you could include it in your repo.


stefan-jansen:
Thank you!

PR #217, opened 2022-01-17
Update date and participants scraper

fabbro-rada:
Update "sa_selenium.py" to account for changes in the HTML tree of seekingalpha.com

In particular:
- update earnings call pages link
- update procedure to get date
- update procedure to get participants


stefan-jansen:
Thank you, but I"m afraid the SA bot checks don't allow for simple scraping anymore.

PR #274, opened 2022-12-22
README update

calvinceecee:
Just wanted to play around with some of your notebooks, particularly the budgeting section.

Thanks for this.


stefan-jansen:
Thank you, but the first edition is rather outdated at this point so not accepting any PRs for it.

PR #325, opened 2024-08-15
Versioning updates

0xrushi:
- Added dockerfile
- Added pyproject.toml, since poetry has better dependency resolution than pip or conda
- Added dockerfile for x86 as well as m1 mac
- Tested on notebooks till ch7, no issues


stefan-jansen:
Thank you for your contribution. 

I would caution against attempting to install all packages for all notebooks in one environment. This makes the process unnecessarily complicated. Just because there are a lot of notebooks in one repo doesn't mean they are a 'library' to install with the standard packaging tools. 

The content differs from many other repos in that it relies on a broader range of libraries. Hence my recommendation to install the libraries you need as you go along. Not everybody is running all notebooks, or in the same order, etc.

stefan-jansen:

PR #335, opened 2025-09-04
Add environment.yml for CI compatibility

ninjamini325:
This PR adds an environment.yml file in the root directory to fix the CI error where the file was not found. The file contains the same dependencies as installation/ml4t-base.yml.


stefan-jansen:

PR #367, opened 2026-07-09
Add optional FXMacroData macro downloader

roberttidball:
## Summary
- add an optional FXMacroData macro downloader beside the existing FRED macro downloader
- support announcements, calendars, catalogues, COT, FX, commodities, sessions, and risk sentiment
- document example commands for writing parquet outputs under the macro data directory

## Validation
- compiled data/macro/fxmacrodata_download.py from the fork branch content

### DIFF
--- a/data/macro/README.md
+++ b/data/macro/README.md
@@ -81,3 +81,28 @@ to match the book's canonical schema.
 - **Ch8**: `04_fundamentals_macro_calendar.py`.
 - **Ch16**: `01_backtest_first_principles.py`, `06_framework_parity.py`.
 - **Ch18**: `02_spread_estimation.py`, `03_market_impact_calibration.py`.
+
+## Optional FXMacroData Download
+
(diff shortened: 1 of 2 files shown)


stefan-jansen:
Thanks for proposing this. We reviewed FXMacroData and agree it is a useful source for FX macro research, especially for point-in-time announcements, calendars, COT positioning, commodities, sessions, and risk sentiment.\n\nRather than merge a standalone book downloader, we implemented this as a first-class ml4t-data provider so the book and future case studies can use the same maintained library surface. The implementation is in ml4t/data commit 8b2b9c1 (feat: add fxmacrodata provider). It exposes FXMacroDataProvider with endpoint-specific methods and preserves response metadata/data-quality context via include_metadata=True.\n\nClosing this PR because the capability is moving into ml4t-data instead of being added directly to the book repository. The book can add a thin consumer/downloader later that imports FXMacroDataProvider from ml4t-data.

PR #378, opened 2026-07-17
Enforce sealed-holdout boundary in feature selection; fix Viterbi look-ahead in regime features

SatishRockzz:
## Summary

This PR fixes protocol-consistency leaks found during a systematic audit. The repo's methodological infrastructure (purged walk-forward CV, HAC ICs, FDR corrections, the sealed-holdout rule in 06/02) is excellent — these fixes close the places where individual notebooks drift from the protocol the book itself teaches.

### Fixes

1. **Sealed holdout used in feature selection** — `08_financial_features/05_feature_selection.py`, `06_robustness_sensitivity.py`, and `case_studies/etfs/03_financial_features.py` §6 computed IC rankings, BH-FDR discoveries, stability selection, and importances on the full sample including the sealed holdout (`evaluation.holdout_start` in `setup.yaml`), then persisted the selected features for downstream chapters. Since feature selection is a development decision, any model later evaluated on the holdout inherits an optimistic bias. All three now read `holdout_start` from `setup.yaml` (mirroring the exact idiom `02_labels.py` already uses) and filter `timestamp < holdout_start` before any selection computation — labels built from the filtered prices can no longer span into the holdout.

2. **Viterbi decode mislabeled as filtered** — `09_model_based_features/13_regime_as_feature.py` derived hard regimes from `hmm_model.predict()` (full-sequence Viterbi, which conditions on future observations) under a comment claiming filtered/causal probabilities, and routed *test* samples to per-regime experts with those states. `11_hmm_regimes.py` explicitly warns against exactly this. Hard states are now the argmax of the filtered probabilities, with the comment corrected; the vol-based state relabeling flips states and probs consistently.

3. **Missing purge gap** — same notebook: the target is a 5-day forward return but `TimeSeriesSplit` had no `gap`, so the last 4 training labels at each fold boundary contained test-window returns. Now `gap=5`, with a comment linking the purge rationale in 06/02.

4. **Full-sample HMM artifact made self-describing** — `11_hmm_regimes.py`'s saved regime features (EM parameters fit on the full sample, disclosed in prose) renamed to `regime_states_fullsample_demo.parquet` with an explicit contamination caveat pointing to the clean per-fold implementation in `case_studies/etfs/04`. Repo-wide grep confirmed no readers of the old name.

5. **Path signatures** — removed the dormant full-sample volume z-score from the featurizer (the natural extension point where it would silently leak) with a causal-normalization warning; the 70/15/15 split now purges `window + horizon − 1` samples at both boundaries.

### Tests

New `tests/test_holdout_boundary.py` statically guards that the three fixed notebooks source the boundary from `setup.yaml` and filter before selection (4 passed, 1 by-design skip when the artifact isn't generated locally). Full suite: 437 passed; the one fix-related failure is `test_notebook_sync` correctly flagging the four edited `.py` files as newer than their stamped `.ipynb` twins — **the twins need a re-execution with production data to refresh provenance**, which I couldn't do locally. Happy to adjust if you'd prefer the twins regenerated differently. Hardcoded result narration (e.g., test IC values) will shift slightly on the next production run now that splits are purged; I didn't hand-edit any numbers.

### DIFF
--- a/08_financial_features/05_feature_selection.py
+++ b/08_financial_features/05_feature_selection.py
@@ -57,12 +57,14 @@
 """Feature Selection and Deduplication — reduce feature candidates to a focused production-ready set."""
 
 import warnings
+from datetime import date
 
 import matplotlib.pyplot as plt
 import numpy as np
 import polars as pl
 import seaborn as sns
 import statsmodels.api as sm
+import yaml
 from ml4t.diagnostic.metrics import pooled_ic
 from scipy.cluster.hierarchy import fcluster, leaves_list, linkage
 from scipy.spatial.distance import squareform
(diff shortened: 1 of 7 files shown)


stefan-jansen:
Thanks @SatishRockzz — we re-executed the affected notebooks in production and adopted your sealed-holdout boundary and regime look-ahead fixes in #379 (now merged). Much appreciated.

PR #426, opened 2026-07-28
Gate notebook provenance at the merge, over the change's own notebooks

stefan-jansen:
## What

The notebook provenance check moves from a pre-commit hook into the `guards` CI job, scoped to the notebooks the change actually touches.

## Why

The hook ran with `always_run: true` and `pass_filenames: false`, so **every local commit scanned every notebook in the repo**. One stale notebook anywhere blocked every commit by every agent, including commits that touched no notebook at all.

Four agents are currently working four case studies against a red CI gate, and a verified fix (`etfs::08_tabular_dl`) sat uncommitted for a day because of this. Uncommitted work is how work gets lost — a reboot on 2026-07-27 destroyed nine hours of it in another worktree.

## What does not change

The invariant. A committed `.ipynb` must still be its current `.py`, executed in a production run. Only *where* it is enforced changes: a stale render reaches a reader through a merge, not through a local commit.

## Scope

A change owns a notebook if it edited the notebook, or edited the paired `.py` — changing the `.py` is exactly what makes the render stale.

The diff form differs by event, because the question differs:

- **pull request** — "what does this branch add on top of the base": merge-base diff, `origin/<base>...HEAD`. Commits that landed on the base meanwhile are not this branch's problem.
- **push to main** — "what does the published tree become": two-dot diff against the previous tip, `<before>..HEAD`. A merge-base diff here would let a force-push revert a notebook relative to the published tip without the gate seeing it.
- **previous tip unavailable** (first push, or a force-push that dropped it) — no honest narrow scope exists, so every notebook is checked. `main` is currently 0 stale and 0 test-mode, so this passes rather than wedging.

The gate runs on push as well as on the PR. Branch protection means everything should arrive via a PR, but this workflow's own header records exports landing on `main` as pushes with nothing CI ran having seen them — a gate covering only the path we believe is the only path is how that happened.

## Verification

Adding one line to `case_studies/etfs/03_financial_features.py` scopes the check to that single notebook and exits 1:

```
checking 1 notebook(s) changed relative to origin/main:
  case_studies/etfs/03_financial_features.ipynb

STALE (paired .py changed since the notebook was executed — re-run in the canonical env):
  case_studies/etfs/03_financial_features.ipynb
```

`--since` with no changed notebooks exits 0. Both diff forms were exercised against a real range. The whole-repo `check` and `check --strict` behave exactly as before.

`guards` gains `fetch-depth: 0` because the diff needs history, not a shallow tip.

## Review history

Roborev blocked the first two pushes, both on real defects in this change: the gate firing only on `pull_request`, and the triple-dot diff on pushes. Both are fixed here (jobs 9509, 9512).

### DIFF
--- a/.github/scripts/notebook_provenance.py
+++ b/.github/scripts/notebook_provenance.py
@@ -94,15 +94,24 @@ def iter_notebooks() -> list[Path]:
     return sorted(out)
 
 
-def notebooks_changed_since(ref: str) -> list[Path]:
-    """Notebooks this branch is answerable for, relative to ``ref``.
+def notebooks_changed_since(ref: str, merge_base: bool = True) -> list[Path]:
+    """Notebooks this change is answerable for, relative to ``ref``.
 
-    A branch owns a notebook if it edited the notebook OR edited the paired
+    A change owns a notebook if it edited the notebook OR edited the paired
     ``.py``, since changing the ``.py`` is exactly what makes the rendered
-    notebook stale. Notebooks nobody on this branch touched are somebody else's.
+    notebook stale. Notebooks nobody touched are somebody else's.
+
+    ``merge_base`` picks which question is being asked. For a pull request it is
+    "what does this branch add on top of the base", so the diff runs from the
+    merge base (``ref...HEAD``) and commits that landed on the base meanwhile are
+    not this branch's problem. For a push it is "what does the published tree
+    become", so the diff must run against the previous tip itself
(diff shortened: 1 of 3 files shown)


stefan-jansen:
Closing in favour of the group model: this is a 3-file infra PR that took five pre-push gate rounds and never landed, and it has gone conflicting again as main moved. The six commits are preserved on the branch and in ~/ml4t/public-provenance; the work will land as part of an infra group rather than on its own.

PR #442, opened 2026-07-31
Stage 04 model-based features: fold routes, IC scope, and a producer blocked by its own output

stefan-jansen:
Three of the nine case studies, stage 04 (`04_model_based_features`). Each is re-run end to end on production data at full surface, and each `.ipynb` carries a provenance stamp recording that.

## crypto_perps_funding could not run at all

The notebook called `load_modeling_dataset` before its write. That loader reads `features/model_based.parquet` — the file this notebook produces — and validates it against the fold geometry derived from the current label frame. The shipped artifact dated from May and failed outright (fold 0 validation coverage 0 of 1091 dates), so the producer was blocked by its own stale output with no way to regenerate it.

The loader call now runs **after** the write, where it is a real check rather than a circular one: the artifact just written has to assemble through the same path every downstream model notebook takes, and the financial columns that come back have to equal the frame read at the top.

## The fold route

All three derived folds from a price or feature frame; the consumer derives them from the label frame, and the two indexes differ because a forward return needs a window to resolve. `fx_pairs` was worst: it replayed a stored `config/cv_config.json` that numbers folds **oldest-first** while `generate_cv_splits` numbers them newest-first, so `fold = 0` named opposite windows on the two sides of the join. The shipped fx artifact does carry the canonical direction — there is a `model_based.parquet.prefoldremap.bak` beside it — so the notebook did not produce what is published.

All three now use `generate_cv_splits` on the label frame, with the buffer read from `setup.yaml` instead of a literal.

## The IC screens

`cme_futures` reported a screen as out-of-sample that took a whole fold, training dates included, and cut the holdout at a literal `2024-01-01` applied to the decision date. It now takes each fold's validation window only, and the seal binds on the outcome endpoint: a decision carrying a 5-session forward return is realized five sessions later, so the last usable decision date is 2023-12-21, derived from `setup.yaml` and the traded calendar.

All three fed `compute_ic_hac_stats` a series that was never sorted. It reads row order as time order and does not sort, and `partition_by` returns groups in frame order — so the Newey-West correction ran over an arbitrary permutation of the timeline, not even stable across runs. All three sort first now, and the HAC lag comes from the label horizon rather than a bare constant.

## An HMM that could not converge

`fx_pairs` fitted its regime HMM on native decimal FX returns. A daily FX log return has a standard deviation near 0.005, so its variance is about 2.4e-5 against `GaussianHMM`'s `min_covar` floor of 1e-3: the regularizer set the covariance, not the data, and the shortest fold had no stable fit at all. Scaled to percent, all eight folds converge (best log-likelihood 568 to 1146).

## Numbers that moved

Expected — these notebooks' folds were wrong.

| | before | after |
|---|---|---|
| crypto `garch_cond_vol` IC | -0.0396 | -0.0415 |
| crypto `garch_vol_zscore` IC | -0.0039 | -0.0076 |
| crypto `garch_asymmetry` IC | +0.0043 | +0.0048 |
| crypto mean stress prob (per fold) | 16.4% / 22.0% | 30.5% / 4.5% |

The stress probabilities move furthest because the fold windows themselves changed.

## Also

`utils/artifact_specs.py` gains `load_setup_config`: `load_evaluation_config` returns only the `evaluation` section, and callers needing the label buffer were re-parsing the YAML by hand. `cme` dropped a dead `apply_ml4t_style` call and two prose passages asserting coverage figures no cell computed. `fx` moved its result numbers out of `display(Markdown)` into a tagged results cell and off two figure titles that interpolated a computed value.

## Not in this PR

The other five case studies at this stage carry evidenced defects that were diagnosed and then reverted unexecuted, because their re-runs did not fit and the `notebook-sync` provenance gate correctly refuses edited code carrying outputs from the old code. They are written up with locations and evidence in the tracker, including one HIGH: an IC screen that runs over training bars and a sealed holdout.


stefan-jansen:
Correction to the PR description, and two follow-ups from review.

**My arithmetic was wrong.** The description says the HMM observation variance sits "about 2.4e-5 ... about four hundred times smaller" than `GaussianHMM`'s `min_covar` floor of 1e-3. A daily FX log return with standard deviation near 0.005 has variance 2.5e-5, and 1e-3 / 2.5e-5 is **40**, not 400. The direction of the argument is unchanged — the variance is still an order of magnitude below the floor, the regularizer still sets the covariance, and the shortest fold still had no stable fit — but the factor was off by ten.

Rather than correct one asserted number with another, the notebook now **measures** it: the cell prints the native variance, the floor, the ratio, and the variance after scaling. Nothing about that trade-off is asserted in prose any more.

**Also fixed here:** under a reduced run (`MAX_SYMBOLS > 0`, i.e. CI) `crypto_perps_funding` picked its universe by row count in two places independently — once in the notebook and once inside `load_modeling_dataset` — with no tie-break. Row counts tie readily on that panel, so the two selections could differ and the post-write assembly assertion could fail for reasons unrelated to the artifact. The notebook now breaks ties on the symbol name and passes the resolved universe to the loader explicitly, so the two cannot disagree.

**Filed, not fixed here:** `fx_pairs/05_evaluation.py` still selects validation rows by joining a fold id from the frozen `config/cv_config.json` against the artifact's `fold` column, and the two number folds in opposite directions — fold 0's config window (2015-10-28..2016-11-14) and fold 0's artifact span (2018-01-04..2023-12-28) do not intersect, so that filter returns zero rows. This predates this PR: the shipped artifact already carried the canonical direction from a manual remap. It belongs to the evaluation stage, not this one, and is tracked separately.

Both notebooks are re-run and re-stamped after these changes.

PR #451, opened 2026-07-31
Stage 01 etfs: five figures instead of two tables, and a cost line per fund

stefan-jansen:
Rebuilds `case_studies/etfs/01_feasibility_analysis` against `rules/stages/01-feasibility.md`, following `case_studies/cme_futures/01_feasibility_analysis` as the reference shape.

The notebook goes from 936 lines and 2 figures to 400 lines and the 5 figures the stage requires. The lines came out of the hand-rolled distribution statistics, the re-printed exceedance table, the fold date table, and prose that restated what a figure carries.

## What changed

**Cost is computed, not typed.** The old notebook assumed a flat 20 bps round trip in prose while `setup.yaml::costs` declares a per-share commission plus a half-spread in dollars per share. Both are now read from the config and converted at the fund's own price. Each move is scaled by what its own fund charged at the price the position would have opened at, so break-even sits at one on an axis every fund shares. The per-fund chart summarises that at each fund's median close; the scaling does not, because a per-share cost in bps depends on the price at the time and an 18-year median is not known then.

**The splitter owns the folds.** The hand-rolled fold loop and its hardcoded 10-year and 1-year train and validation literals are replaced by `generate_cv_splits`, which reads `setup.yaml::evaluation` and is the same call `04_model_based_features` makes. The `splits_df` date table is replaced by `fold_timeline`, which draws the boundaries the splitter returned rather than recomputing them.

**Section B is sealed.** The old notebook built its return distributions, cost exceedance and feasibility ratio across 2006-2025, holdout included. Everything in B now computes on the development window.

**Breadth is counted on the dates the strategy acts.** Eligible funds per month-end decision date, against the largest position count the Chapter 16 sweep asks for, with the dates on which nothing is eligible kept rather than dropped by an inner join.

**Signal persistence is new.** `panel_acf` gives the within-fund autocorrelation of the month-to-month return with a white-noise band, computed inside each fund rather than over the stacked panel. It is presented as serial correlation of the carrier and nothing more.

**No verdict.** The `PROCEED` / `KILL -- edge too thin` assessment is gone, and the unsigned move is no longer called an edge. Section F states a decision rule per knob instead: the evidence that motivates it and the condition under which a reader would revise it.

**Parameters are declared once.** `START_DATE` and `ADV_THRESHOLD` were both re-assigned further down, which silently killed the papermill override.

**Artifacts.** `config/exploration/feasibility_report.json` is no longer written; a repo-wide search finds no reader. `eligibility.csv`, which `02_labels` and `03_financial_features` do read, is unchanged - the re-run reproduces it byte for byte, so no fixture regeneration is needed.

**Test overrides.** `MAX_SYMBOLS`, `N_SPLITS`, `TRAIN_YEARS` and `TEST_YEARS` named nothing in the notebook; `START_DATE` and `HOLDOUT_START` were injected and then overwritten. `ADV_THRESHOLD` is the only live one and is kept. The CI fixture - 24 funds over the full history - is the reduction, and the declared eight folds still fit it.

## Numbers that moved

Nothing in this stage reaches `run_log/registry.db`; it registers no results.

| | Before | After | Cause |
|---|---|---|---|
| Round-trip cost | assumed flat 20 bps | 0.90 to 38.91 bps per fund, median 9.45 | read from `setup.yaml::costs` and divided by each fund's own price |
| Move-to-cost ratio | 14.2x | 30x, with 0.967 of moves clearing their entry cost | the median cost fell from the assumed 20 bps to the computed 9.45, and the window is now the development window |
| Breadth | 88 to 96 per fold test period (prose said "70+") | 0 to 96 across 216 decision dates, under the floor on the 12 dates of 2006 | counted per decision date rather than per fold test year, and dates with nothing eligible are kept |

The median absolute 21-session move is unchanged at 287.2 bps.

## Verification

```
$ uv run python scripts/check_notebook_conformance.py --stage 01 --case-study etfs -v
=== stage 01 (01-feasibility.md): 200-400 lines, 5+ figures, 9 required sections ===
  PASS  etfs
```

- `tests/test_case_studies.py -k "etfs and 01_feasibility"`: 1 passed
- `tests/test_notebook_sync.py`: 19 passed
- Executed on production data (100 funds, 420,462 rows), 5 figures, 0 errors, 0 unexecuted cells, provenance stamped
- Every figure extracted from the executed notebook and checked by eye against its title

## Known limitations recorded in the notebook

`close` is adjusted for splits and distributions, so early prices sit below what a fund traded at: dollar volume is understated there, and the round trip, being dollars per share over a price, is overstated. Correcting it needs an unadjusted price the dataset does not carry, and would move `eligibility.csv`, which two downstream notebooks and the fixture repo read.


stefan-jansen:
CI: the four required checks pass - `changes`, `guards`, `lint`, `test-unit`.

The `cs-etfs` job is red, and it is red on `main` with the identical set: `03_financial_features`, `08_tabular_dl`, `11a_pca`, `11b_ipca` and `18_strategy_analysis` fail on both, with the same five errors (`5 failed, 14 passed, 4 skipped` either way). `01_feasibility_analysis` passes on both. This branch neither adds a failure nor fixes one; the five are recorded separately with their errors.

Measured against run 30604214957 job 91073225731 on `main` and run 30604661492 job 91074630041 on this branch.

PR #459, opened 2026-07-31
Rebuild the sp500_equity_option_analytics labels on a tradable, sealed, null-guarded basis

stefan-jansen:
Brings `case_studies/sp500_equity_option_analytics/02_labels` to the stage-02 standard and
re-runs it on production data. Three defects here move a shipped label file, so every number
in the notebook and every classification metric and backtest built on these labels changes.

## The labels were measured on a price series that is not tradable

`load_sp500_daily_bars` returns the printed `open` and `close` together with `adj_factor`, the
cumulative factor that makes a price comparable with the rest of that security's history. The
notebook used the printed prices, so `close[t+h] / open[t+1] - 1` recorded corporate actions as
returns: GE's July 2021 reverse split produced a **+683%** weekly label, XRX's 2017 spin-off
**+311%**, and AAPL's four-for-one subdivision **-74%**. 4,251 weekly labels differed from the
adjusted one by more than a percentage point.

The entity was also wrong. `adj_factor` restarts when a ticker is reassigned to another
company, and fifteen tickers here cover two securities each, thirteen of them on consecutive
sessions. The label now rides `close * adj_factor` inside a `sec_id`, which is what
`case_studies/sp500_options/_underlying_returns.py` already does on this dataset.

| | before | after |
|---|---|---|
| `fwd_ret_5d` kurtosis | 2249.1 | 16.48 |
| `fwd_ret_5d` range | -0.886 to +6.828 | -0.667 to +1.112 |
| `fwd_ret_10d` kurtosis | 1474.3 | 10.09 |

Verified while reviewing: `adj_factor` carries **cash dividends as well as splits**, so the
label is a total return to a holder of the share and the notebook says so. 8,096 of the 8,098
`CashDiv` rows move it - CB's 2017-03-29 ex-date carries `adjustment_factor` 0.995 against a
0.50% dividend yield. The loader docstring at `data/equities/loader.py:299` calls it a factor
"for split adjustment", which is what made this worth checking twice.

## The direction labels fabricated their tail

`pl.when(fwd_ret > 0).then(1).otherwise(0)` has no null guard, and a null predicate in Polars
is not true, so `.otherwise(0)` fired wherever the return was null. The last five and ten rows
of every security shipped a confident "down" with no forward window at all, and the
`.drop_nulls()` on the write could not remove them because the value was not null. Both files
now null where their return is null.

| | before | after |
|---|---|---|
| `fwd_dir_5d` rows | 635,703 | 632,522 (3,181 fabricated) |
| `fwd_dir_10d` rows | 635,703 | 629,378 (6,325 fabricated) |
| `fwd_dir_5d` positive share | 54.02% | 54.61% |
| `fwd_dir_10d` positive share | 55.24% | 56.34% |

Section D asserts this directly, so it cannot come back silently, and the validity figure shows
each direction label nulling exactly where its return does.

## Nothing was sealed against the holdout

`HOLDOUT_START` was bound at line 82 and never referenced again, so the distribution, the annual
breakdown, the one figure, the baseline IC and the autocorrelation all included the 2021
holdout, and the takeaways quoted a t-statistic computed on it. Every diagnostic from Section D
on now runs on the development window, sealed on **each label's own endpoint**: the timestamp
shifted forward by that label's horizon inside the security, then filtered. A signal-date filter
would have kept the last five sessions of 2020, whose outcomes resolve in the holdout - 2,523
rows of the primary label.

The notebook adds itself to both lists in `tests/test_holdout_boundary.py`, so the seal is now
checked mechanically. It is deliberately absent from `PER_SYMBOL_ENDPOINT_NOTEBOOKS`, whose
regex requires the endpoint to be shifted within `symbol`; here that would be the under-purge
the list exists to prevent, and the comment there says so.

## Two more, found by the pre-push review and confirmed on the data

- The scaled label's denominator took `pct_change()` over consecutive **rows**. Seven securities
  miss sessions inside their own series - LHX and APTV for a month each - so the first return
  after a gap was a multi-session move recorded as one day and it stayed in the twenty-day
  volatility window for twenty rows. `fwd_ret_risk_adj_5d` loses 120 rows and its development
  standard deviation moves 0.16625 -> 0.16627.
- The baseline signal was lagged one **row of the option surface**. Coverage swings on a monthly
  cycle and 14,964 of 414,520 quoted rows have their previous quoted row more than one session
  back, at a median of fourteen, so a fortnight-old volatility was being scored as yesterday's.
  The surface is now joined onto the dense price panel and lagged one session inside the
  security.

## What else the rebuild changes

- **The A-H skeleton**, none of whose eight sections existed. Section D - window validity - did
  not exist in any of the nine and is where three of the filed stage-02 defects surface.
- **Four figures against one.** Validity at the window boundary, the two horizons on identical
  bins, annual dispersion and positive-class share, and overlap decay across the panel. Every
  title is a static claim; the shipped annual-dispersion title interpolated its peak year at
  render time.
- **Every horizon and boundary bound from `setup.yaml`.** The notebook hardcoded its horizons,
  its holdout and its dates, and re-assigned `START_DATE` and `END_DATE` below the parameters
  cell, which killed the papermill override.
- **Library calls replace hand-rolled statistics**: `cross_sectional_ic_series` and
  `compute_ic_hac_stats` for the baseline, `calculate_label_uniqueness` through
  `effective_sample_size` for `N_eff`, `panel_autocorrelation` across the panel rather than a
  lag-1 scalar. The notebook's own prose said HAC standard errors were necessary and then
  computed the naive statistic anyway and quoted it in the takeaways as `t = -1.86`; the naive
  and the HAC statistic are now both reported.
- **The baseline signal is the one the hypothesis names** - the lagged thirty-day at-the-money
  volatility `01_feasibility_analysis` measured the persistence of - rather than 21-day price
  momentum. Mean IC -0.0067, HAC t -0.47 on 6 Bartlett lags, naive t -0.87, over 1,001 sessions.
- **A `.digest.json` sidecar per label file**, recording the content digest of what was written
  and the digest of the price data it was built from.
- **`cv_config.json` is deleted.** Per the stage ruling, stage 02 writes none. This case study
  was the only one with live readers: `05_evaluation` took `test_start` and `n_splits` from it,
  both identical to `setup.yaml` (2021-01-01 and 2), and both now come from
  `load_evaluation_config`. `05_evaluation` is re-executed on production data because its
  feature evaluation scores against the corrected `fwd_ret_5d.parquet`.

## Every artifact that moved

| file | rows before | rows after | digest |
|---|---|---|---|
| `fwd_ret_5d.parquet` | 632,508 | 632,522 | `7280bba2710cfa08` |
| `fwd_ret_10d.parquet` | 629,343 | 629,378 | `1f02b172e1edf0b8` |
| `fwd_ret_risk_adj_5d.parquet` | 619,887 | 620,010 | `4b2040f93c8e59f7` -> `ab5f2381de72fc3f` |
| `fwd_dir_5d.parquet` | 635,703 | 632,522 | `e9bc4b44f7f8c3d8` |
| `fwd_dir_10d.parquet` | 635,703 | 629,378 | `0a15949f45d521df` |

Everything trained on these labels needs re-running.

## Checks

```
$ python3 scripts/check_notebook_conformance.py --stage 02 \
      --case-study sp500_equity_option_analytics -v
=== stage 02 (02-labels.md): at most 330 code lines, 4+ figures, 11 required sections ===
  PASS  sp500_equity_option_analytics
```

`check_notebook_prose.py` clean. `tests/test_holdout_boundary.py` 29 passed, 2 skipped.
`tests/test_notebook_sync.py` 19 passed. `test_case_study_pipeline` green for `02_labels`,
`03_financial_features` and `05_evaluation`. Every figure was extracted from the executed
notebook and read against its title.

`test_no_machine_specific_paths_in_committed_notebooks` is red on `main` before this branch and
after it, on roughly thirty notebooks; it is tracked separately and nothing here adds to it.

## Left out of scope

`03_financial_features` builds `mom_21d` and `rv_20` from the same unadjusted closes, its
`cv_config.json` block now tells the reader to run a notebook that does not write the file,
`05_evaluation`'s fold-stability branch reads a `splits` key no producer ever wrote, and
`case_studies/utils/backtest_loaders.py` gives this case study a price panel built from the
printed `close` with `adj_factor` explicitly dropped - so the label and the backtest now measure
returns on two different price series. All four are filed with their evidence.


stefan-jansen:
Carried onto the group A integration branch `group/csA-features`, which collapses the four stage-02 label branches into one. This PR's commits are on it as cc38f07e, 87731666, 5cbc4986, with the case study's files byte-identical to this branch.

PR #461, opened 2026-07-31
Rebuild the sp500_options label notebook, and seal it on the settlement date

stefan-jansen:
`case_studies/sp500_options/02_labels` rebuilt to the stage-02 standard, re-run on
production data, and sealed on the date each position settles.

## The leak

The primary label `ret_to_expiry` is a short straddle carried to expiration, so the
date its outcome is known is written into the contract rather than fixed at a number
of sessions after the signal. The notebook sealed its baseline IC on the **signal**
date, so every trade opened from about 2020-12 onward settled inside the sealed 2021
holdout and still entered the statistic. Everything else was worse: the `primary`
frame was the unsealed full sample, so the mean, the standard deviation, the hit
rate, the autocorrelation, the effective sample size and the distribution figure all
included the holdout.

`_label_end` is now the date each label's outcome is fully observed - the contract's
`expiration` for the primary label, the recorded exit session for the four
fixed-horizon variants - and sections D through H are scoped on it. That is the same
boundary `04_model_based_features`, `05_evaluation`, `06_linear` and `07_gbm` already
derive from the `dte_calendar` column this notebook writes.

## Every number that moved, and why

Three of the five label parquets reproduce **bit for bit** - `ret_to_expiry`,
`fwd_ret_5d` and `fwd_ret_10d`, checked row by row against the previous construction on
the production artifacts. The two delta-hedged labels lose rows, for the reason in "The
hedge that was not a hedge" below. Everything else that moved is the notebook's own
diagnostics.

| Quantity | Was | Now | Cause |
|---|---|---|---|
| Primary label mean | -0.0242 | -0.0487 | was the unsealed full sample; now the development window |
| Primary label std | 1.1524 | 1.0369 | same |
| Share of trades profitable | 59.4% | 58.5% | same |
| Rows the diagnostics see | 354,473 | 282,470 | the endpoint seal drops 5,752 rows the signal-date seal kept, on top of the 66,251 the old frame never sealed at all |
| `N_eff` | 49,590 | 22,897 | the old figure came from `1 + 2·Σ max(ρ_k, 0)` on a per-symbol Spearman ACF over the full sample; the new one is Chapter 7.2's average uniqueness via `calculate_label_uniqueness`, weighting each trade by its own exposure |
| Lag-1 autocorrelation | +0.742 | +0.915 | was the mean of per-symbol Spearman correlations; now the panel Pearson autocorrelation, demeaned within symbol and paired on the panel calendar |
| Baseline mean IC | -0.0108 | -0.0100 | see below |
| Baseline HAC t | -1.23 (p 0.220) | -1.06 (p 0.291) | see below |
| Sessions scored | 986 | 968 | the endpoint seal |
| Minimum cross-section | 3 | 123 | half the median cross-section of 246, instead of a bare integer |
| `fwd_ret_dh_5d` rows | 352,179 | 349,934 | an incomplete hedge path now nulls the label |
| `fwd_ret_dh_10d` rows | 333,102 | 323,847 | same |
| `fwd_ret_dh_10d` mean / std (development) | - | +0.0159 / 0.2182 | as above; both were previously computed on the unsealed sample |

**The baseline IC, decomposed.** Holding the minimum cross-section at the old value
of 3 isolates the seal: signal-date seal -0.0108 with HAC t -1.23 over 986 sessions
(reproducing the shipped value exactly), endpoint seal -0.0112 with HAC t -1.26 over
968. Raising the minimum cross-section to half the median then moves it to -0.0100
with HAC t -1.09, and widening the Newey-West bandwidth from the median window to the
longest one (20 to 24 sessions of exposure, 23 Bartlett lags) gives -1.06. So the seal
made the floor slightly *more* negative and the
cross-section rule pulled it back; neither changes the reading, which is that the
variance risk premium points the opposite way to the hypothesis and the data cannot
separate the estimate from zero. The naive t of -3.21 against the HAC t of -1.09 is
the overlap being counted as evidence, and the notebook now says so.

## The label set against its own contract

`setup.yaml: labels.primary` plus `labels.variant_buffers` is now exactly what gets
written, five files:

- **`fwd_ret_exec_{5,10}d` deleted.** They appeared exactly once in the repository -
  on the line that built them. The writes and the two committed parquets are gone.
- **`fwd_ret_{5,10}d` and `fwd_ret_dh_{5,10}d` declared** in `labels.variant_buffers`,
  which previously listed only `fwd_ret_dh_10d`. They are written and read
  (`03_financial_features`, `05_evaluation`, `90_ic_diagnostic`,
  `13_dl_time_series/dl_sequences.py`), so the declaration had to catch up with the
  code rather than the other way round.
- **Not moved into `labels.variants`.** That list drives the sweep - it is what
  `tests/test_sweep_config_seam.py` reconciles against `backtest.sweep.top_k_grid`,
  and what the latent-factor and Ch20 paths iterate - and these four were removed
  from the sweep on 2026-05-17 because the vectorized backtest path treats their
  multi-day returns as daily ones. `variant_buffers` already existed for exactly this
  case and is not read by the sweep, so it is where they belong.

## The hedge that was not a hedge

`fwd_ret_dh_{5,10}d` add the hedge P&L accrued along the contract's own delta path. The
sum ignored days the contract was not quoted on, and a wholly missing path was turned
into zero by a `fill_null(0)`, so a partially hedged - or entirely unhedged - trade was
written out as a delta-hedged return. Measured on the production artifacts: 2,245 rows
at five sessions and 9,255 at ten, 0.6% and 2.8% of each label, of which 16 and 5 had no
hedge path at all.

`accrued_hedge_pnl` now returns the number of days it found a quote on alongside the
accrued P&L, and a label whose path is short of its horizon is null. That is a fourth
row in the Section D reconciliation rather than a silent pass.

Both labels are diagnostic-only - neither is in the sweep - but `fwd_ret_dh_10d` is
read by `03_financial_features`, `05_evaluation`, `90_ic_diagnostic` and
`dl_sequences.py`, so those four will show different numbers the next time they run.

## Two things that were not reproducible

**The digest of the delta-hedged labels moved on every re-run.** Three executions of
identical code on identical inputs produced three different content digests, while the
other three labels reproduced exactly. The cause is the parallel sum over each holding
path: an unordered reduction re-associates the floating point, which moved the label by
up to 9e-16 - immaterial to any number, and enough to make the sidecar useless as a
content address. The accrual is now summed in day order and all five digests reproduce
across processes.

**A reduced run poisoned the cache for the next full one.** `ensure_label_artifacts`
returned whatever was on disk, so a `MAX_SYMBOLS` or `START_DATE` run wrote a reduced
`contract_returns.parquet` that the next default run then accepted as the full panel.
The requested scope is now recorded beside the artifact and the cache is only reused for
the same request.

**And `MAX_SYMBOLS` counted horizons on a symbol-subset calendar.** The builder derived
its session calendar after filtering to the sampled symbols, so any session none of them
was quoted on vanished and the exit dates sat that many sessions further out than
declared, while the notebook counted on the whole panel. Latent rather than breaking: the
selection takes the most-quoted names, and those are quoted on all 1,259 production
sessions and all 1,008 fixture ones, so no session is dropped today. The calendar now
comes from the whole panel and the reduction thins only the entries it is applied to. A
run at `MAX_SYMBOLS=5, START_DATE=2018-01-01` executes clean, both parameters bite (5
names, signals from 2018-01-02), and every Section D assertion holds.

## The rest

- **Papermill.** `START_DATE` was declared and never read. Both parameters now reach
  `ensure_label_artifacts`, which no longer hands back a cached full-panel artifact
  when a narrower window is requested - without that the override could not have bitten
  whatever the call site did.
- **`N_eff`** comes from `case_studies/utils/label_diagnostics.effective_sample_size`,
  which gains an optional `horizon_col` so a label whose window is not the same length
  for every row is weighted per row rather than at the median. Checked at `h = 1` on a
  synthetic gapless grid, where it returns `N` exactly, and the scalar and per-row forms
  are checked to agree where the window is constant. The overlap of a trade held `h`
  sessions is `h` return intervals, not `h + 1`: the execution delay shifts the window
  without lengthening it.
- **The dead `results` dict** (60 lines, under a "Results Collection" heading) is gone,
  along with the sentence asserting what a later notebook found.
- **`config/cv_config.json`** was written here and opened by nothing. The write and
  the committed file are gone; section H states positively where the folds come from.
  The 35D cross-validation buffer and the label's own resolution time are separate
  rows of the audit record, because they are separate numbers here.
- **Digest sidecars** for all five labels, keyed
  `["timestamp", "symbol", "instrument_id"]`, with `inputs` naming
  `contract_returns`, `hedge_path` for the delta-hedged pair, and `market_data` for
  the primary's settlement closes.
- **Five figures** replace the printed tables: resolution time, window validity at the
  panel boundary, all five label distributions on one axis with identical bins,
  cross-sectional dispersion through time, and overlap decay across the panel with
  each holding window marked. Every PNG was extracted from the executed notebook and
  checked against its title.
- **Both static holdout gates** in `tests/test_holdout_boundary.py` now list this
  notebook, and `tests/test_sp500_options_label_price_basis.py` was updated to the
  rebuilt notebook: it now runs the notebook's own `panel` assignments rather than a
  transcription of them, and requires the HAC bandwidth to be bound from the observed
  windows instead of being a literal.
- **The case-study README** listed executable-return artifacts and `config/cv_config.json`
  among what this stage writes. It now names the five files and points at `setup.yaml`
  for the folds.

## Verification

```
python3 scripts/check_notebook_conformance.py --stage 02 --case-study sp500_options
  PASS  sp500_options
```

```
uv run pytest tests/test_case_studies.py -k "sp500_options and 02_labels"   1 passed
uv run pytest tests/test_holdout_boundary.py                                29 passed, 2 skipped
```

Executed on production data through `jupyter nbconvert`: zero error outputs, zero
unexecuted cells, five figures, no absolute paths, provenance stamped.

`test_notebook_output_hygiene.py::test_no_machine_specific_paths_in_committed_notebooks`
is red on the base commit for thirty-odd unrelated notebooks and is filed separately;
`02_labels` is not among them.

### DIFF
--- a/case_studies/sp500_options/02_labels.ipynb
+++ b/case_studies/sp500_options/02_labels.ipynb
@@ -3,41 +3,43 @@
   {
    "cell_type": "markdown",
    "id": "1fec74bc",
-   "metadata": {
-   },
+   "metadata": {},
    "source": [
     "# S&P 500 Options: Label Engineering\n",
     "\n",
-    "This notebook constructs labels for the S&P 500 Options case study using\n",
-    "**same-contract returns** built directly from the slim daily straddle panel\n",
-    "and the slim lifecycle-preserving raw straddle slice. Labels are short straddle\n",
-    "returns: sell a 30D ATM straddle at mid, buy back the **same contract** at mid\n",
-    "after h days. Positive return = profitable short vol position before costs.\n",
+    "The label is a short straddle's return, and an option position is not a share position:\n",
+    "the thing sold at entry has to be the thing bought back at exit, and it stops existing\n",
+    "on a date fixed when it is sold. This notebook writes that convention down as a formula,\n",
+    "proves each labelled trade has a complete round trip inside one contract, measures how\n",
+    "much independent information overlapping trades carry, establishes the floor a feature\n",
+    "has to clear, and writes the files stage 03 reads.\n",
     "\n",
-    "**Learning Objectives**:\n",
-    "- Construct instrument-level labels from same-contract exit prices\n",
-    "- Implement delta-hedged returns using the held contract's delta path\n",
-    "- Generate walk-forward CV splits with a buffer that spans the ~30-day horizon\n",
-    "- Establish a VRP baseline IC that engineered features must beat\n",
+    "## Learning objectives\n",
     "\n",
-    "**Book Reference**: Chapter 7, Section 7.2 (Label Engineering)\n",
+    "- Express an option label as a round trip in one contract, and say what each leg costs\n",
+    "- Assert - rather than describe - that every labelled window closes on a real quote\n",
+    "- Seal a diagnostic on the date a position settles, not the date it is opened, when the\n",
+    "  settlement date is a property of the contract rather than a fixed offset\n",
+    "- Price the overlap in a label held for a month and sampled every session\n",
+    "- Establish the floor a feature has to clear, under a standard error that prices in that\n",
+    "  overlap\n",
     "\n",
-    "**Prerequisites**: [`01_feasibility_analysis`](01_feasibility_analysis.ipynb)\n",
+    "## Book reference, prerequisites and artifacts\n",
     "\n",
-    "**Primary label**: `ret_to_expiry` - the hold-to-expiry (HTM) short-straddle\n",
-    "return declared in `setup.yaml::labels.primary`. Every downstream model in this\n",
-    "case study trains on it. The horizon-based variants (`fwd_ret_{5,10}d`, their\n",
-    "delta-hedged and executable counterparts) are also built here to characterise\n",
-    "the pricing and hedging mechanics, but they are diagnostic, not modelled.\n",
+    "Chapter 7, Section 7.2. Reads the daily matched-strike straddle panel and the raw option\n",
+    "chain through `load_sp500_options_straddles()` and `load_sp500_options_straddles_raw()`,\n",
+    "whose coverage [`01_feasibility_analysis`](01_feasibility_analysis.ipynb) establishes,\n",
+    "the underlying closes through `load_sp500_daily_bars()`, and `config/setup.yaml`, which\n",
+    "declares the label set, the cross-validation buffer and the holdout boundary.\n",
     "\n",
-    "**Pricing Convention**: Mid-to-mid labels measure the economic signal.\n",
-    "Executable labels (sell at bid, close at ask) bake in the per-contract\n",
-    "bid-ask spread and reveal that the unconditional 10-day short straddle return\n",
-    "falls to **-9.6%** per trade, even though its mid-to-mid counterpart is +5.7%.\n",
-    "The exit-leg spread is what `ret_to_expiry` avoids: cash settlement at\n",
-    "expiration converts the option to intrinsic value with no market exit. Whether\n",
-    "ML selection plus the O'Donovan-Yu cost-mitigation cascade turns this into a\n",
-    "viable strategy is the subject of Chapter 18 (Section 18.8), not this notebook."
+    "Writes five label files - `labels/ret_to_expiry.parquet`, `labels/fwd_ret_5d.parquet`,\n",
+    "`labels/fwd_ret_10d.parquet`, `labels/fwd_ret_dh_5d.parquet` and\n",
+    "`labels/fwd_ret_dh_10d.parquet` - each with a `.digest.json` sidecar beside it.\n",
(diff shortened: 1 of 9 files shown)


stefan-jansen:
Carried onto the group A integration branch `group/csA-features`, which collapses the four stage-02 label branches into one. This PR's commits are on it as b1a07282, with the case study's files byte-identical to this branch.

PR #462, opened 2026-07-31
us_equities_panel 02_labels: run the forward shift before the eligibility screen, and bring the notebook to the stage-02 standard

stefan-jansen:
Brings `case_studies/us_equities_panel/02_labels` to the stage-02 standard and re-runs it on
production data. `case_studies/cme_futures/02_labels.py` (#445) is the shape this follows.

## The defect that moves the label files

`02_labels.py:154 -> 195-198` ran the forward shift on the eligibility-filtered frame.
`apply_pit_filters` dropped rows priced below $5 or below $1M ADV, and the shifts then counted
the rows that survived, so a stock that fell below either threshold and recovered carried a label
spanning the whole excursion - reported as a 1-, 5- or 21-session return. `adv_21d` is also null
for each symbol's first 20 rows, so every entry into the panel was a gap boundary.

The shift now runs on the complete price series and the screen is applied afterwards. That reaches
the same eligible universe, because the screen removes whole rows and never changes the window of
a row it keeps.

A window is additionally nulled where it spans more calendar time than `h` sessions of a
five-session week plus a week of exchange holidays. The panel holds stocks that stop trading for
months and resume under the same ticker - 86 such windows at h=1, 10,399 at h=21 - and without
that rule the shift closes over the hole and reports a multi-year price change as a one-session
return.

### Measured, on the three shipped artifacts

| Artifact | Rows old -> new | Values changed | Median \|delta\| | Max \|delta\| |
|---|---|---|---|---|
| `fwd_ret_1d` | 9,271,950 -> 9,272,483 | 39,658 (0.43%) | 0.0489 | 92.75 |
| `fwd_ret_5d` | 9,259,265 -> 9,261,926 | 150,825 (1.63%) | 0.0639 | 102.42 |
| `fwd_ret_21d` | 9,208,607 -> 9,215,723 | 448,243 (4.87%) | 0.0788 | 35.04 |

Cause, in both directions: rows gained are anchors whose window previously spanned an
ineligibility gap and were dropped by a later `drop_nulls`; rows lost are windows that span a
genuine hole in the stock's series and are now null. The changed values are the anchors whose
window compressed. The symbol count rises from 3,172 / 3,170 / 3,157 to 3,173 on all three,
because a stock whose eligible rows were too few to span the horizon can now carry a label.

Every registered number in this case study is downstream of these files, and
`case_studies/utils/cv_window.py:119-126` derives the fold boundaries from each label parquet's
own timeline, so the fold geometry moves with them. `fwd_ret_21d` gains one date.

## What else the notebook lost, and why

- **The naive t-statistic on an overlapping-label IC series** (`:349`, `np.std(ddof=0)`) - now
  `cross_sectional_ic_series` into `compute_ic_hac_stats(..., label_horizon=h)`, which reports
  `naive_t_stat` beside the corrected one so the correction is visible rather than merely applied.
- **`N_eff` asserted in prose** (`:209-215`) - now computed by `effective_sample_size`, which is
  `calculate_label_uniqueness` under Chapter 7.2's average-uniqueness definition. The one-session
  label is the check that fixes the arithmetic: with disjoint windows `N_eff` comes back equal to
  `N` exactly, which it does (8,058,076).
- **The hand-rolled autocorrelation loop** (`:366-397`) - now `panel_autocorrelation` across the
  panel rather than five printed lag scalars.
- **The holdout seal.** `timestamp < holdout_start` is not a seal; diagnostics are now scoped on
  `_label_end`, derived by shifting the timestamp by the declared horizon within the symbol, on the
  complete price series - shifting the screened frame would return the horizon-th next *eligible*
  row, which is a later date and depends on eligibility after the decision. The notebook adds itself
  to both lists in `tests/test_holdout_boundary.py`.
- **The `cv_config.json` write** (`:254-274`, `:475-477`), per the stage-02 ruling. Value-neutral
  here: `05_evaluation.py:98-99` loads the committed file and then hand-rebuilds its own folds, and
  the committed `config/cv_config.json` is untouched, so nothing breaks. Filed separately.
- **The dead `results` dict** (`:504-552`), the CV-config verification block (`:404-433`), and the
  symbol-count reconciliation (`:478-496`) that typed four numbers into markdown which the next
  cell printed from the data.
- Horizons, the holdout boundary and the panel end date are bound from `config/setup.yaml`
  instead of typed as literals.

## Structure

The A-H skeleton, and four figures replacing six printed tables:

1. Non-null label rate by position from the end of each stock's series - the null tail is exactly
   as deep as each label's own horizon.
2. All three labels on one axis, identical bins, log count axis, dispersion in the legend.
3. Annual mean of the daily cross-sectional dispersion of the primary label.
4. Panel autocorrelation of all three labels against lag, each decaying to zero at its own horizon.

339 -> 328 code lines; 0 -> 4 figures. Three tagged `results` cells, one each in E, F and G.

## The baseline floor

One signal against the primary label only: the skip-month momentum `setup.yaml` declares as the
causal treatment and `03_financial_features` recomputes as `ret_12m_skip`. Mean IC 0.0119 over
6,314 sealed sessions, HAC t 4.95 on 9 Bartlett lags against a naive 5.60.

It is computed on the complete price series, not on the screened development frame - the same rule
the label follows, for the same reason. `03_financial_features` still shifts its own lookbacks on
the filtered frame, so its `ret_12m_skip` is not yet the quantity this baseline measures; that is
tracked and out of scope here.

## Verification

```
python3 scripts/check_notebook_conformance.py --stage 02 --case-study us_equities_panel   # PASS, 0 failing
uv run pytest tests/test_holdout_boundary.py -k us_equities_panel                          # 3 passed
ML4T_DATA_PATH=~/ml4t/test-data/data ML4T_OUTPUT_DIR=/tmp/ml4t-test-output \
  uv run pytest tests/test_case_studies.py -k "us_equities_panel and 02_labels"             # 1 passed
```

Executed on production data end to end: 0 error outputs, 0 unexecuted cells, 4 figures, provenance
stamped. Each label parquet is written with a `.digest.json` sidecar carrying the digest of the
price frame it was built from - over price *and volume*, since volume decides which rows are
written.

All 22 `us_equities_panel` CI nodes were run together against the fixture: 20 passed, 2 skipped, 0
failed, including `03_financial_features`, `04_model_based_features`, `05_evaluation` and every
model stage that consumes these label files.

## Known, not fixed here

The `$5` floor is applied to `adj_close` rather than to the price that printed, which is a screen
on cumulative later corporate actions. It is unchanged in this PR: `03_financial_features:131`
rebuilds the same screen and inner-joins these labels to the rows it keeps, so correcting it in
`02_labels` alone would leave the two notebooks disagreeing while `03` went on governing the
trainable universe. It is tracked, with the two notebooks scoped together.


stefan-jansen:
Carried onto the group A integration branch `group/csA-features`, which collapses the four stage-02 label branches into one. This PR's commits are on it as 12fe41bd, 040bd71b, b8a1ed8e, with the case study's files byte-identical to this branch.

PR #463, opened 2026-07-31
Rebuild nasdaq100 labels: each label from its own nulls, horizon as a duration

stefan-jansen:
Rebuilds `case_studies/nasdaq100_microstructure/02_labels` to the stage-02 standard and fixes two
defects that change what lands in `labels/*.parquet`.

## The two artifact defects

**1. `fwd_ret_5m` was written from the primary label's null set.** `df.drop_nulls(subset=["fwd_ret_15m"])`
gated the write loop for all four labels, so the 5-minute label lost every row on which the 15-minute
label was null - the last ten bars of every symbol-session. Each label is now written from its own
nulls, and Section D asserts that each session is short by exactly its own horizon and no more.

| Label | Before | After | Change |
|---|---:|---:|---|
| `fwd_ret_5m` | 19,142,250 | **19,652,710** | **+510,460 rows**, +2.67%, all at the session close |
| `fwd_ret_15m` | 19,142,250 | 19,142,250 | unchanged |
| `fwd_ret_60m` | 16,845,180 | 16,845,180 | unchanged |
| `fwd_dir_15m` | 19,142,250 | 19,142,250 | unchanged |

The shipped `fwd_ret_5m` was missing 2.60% of its correct support and ended at 15:44 instead of
15:54. Its distinct-timestamp count goes from 189,375 (505 x 375) to 194,425 (505 x 385), and
`case_studies/utils/cv_window.py:119-126` derives that label's folds from exactly that timeline, so
the fold boundaries for any model trained on `fwd_ret_5m` move with it.

**2. A bar count was used as a time horizon.** `shift(-15)` on a frame nothing had checked is a
15-minute return only where the minute grid is complete. The horizon is now the duration
`setup.yaml` declares, converted into bars against a **measured** grid interval, with the exit
resolved by timestamp (`fixed_time_horizon_labels(horizon=..., tolerance="0s")`) so a missing bar
nulls the label instead of letting the shift reach past it.

**Measured on production: this moves zero rows.** The grid is exactly uniform - 19,908,044
regular-hours bars, all 51,046 symbol-sessions exactly 390 one-minute bars, zero interior gaps. The
only quote dropout is ALXN's final 104-bar session on 2021-07-21 (delisting), which removes a whole
trailing session rather than punching a hole in one. So the two constructions coincided on this data
and the defect was unenforced code rather than wrong values - which is exactly why it needed
asserting rather than trusting. Section D now proves it every run: every labelled window spans
exactly its declared horizon in wall-clock time.

Two prose claims that were wrong independently of the data are gone: that `shift(-15)` on minute bars
and `shift(-1)` on 15-minute bars produce the same return (they span 14 and 15 bars from different
bases), and that the horizons are "5 bars = 5 minutes".

## Other numbers that moved, and why

The baseline IC in Section G is computed differently, so its value moved. The previous cell thinned
the panel to every 15th timestamp (~9k of 135k decision minutes) and took a naive t-statistic on the
result; the standard forbids the naive statistic on an overlapping-label IC series and names
`cross_sectional_ic_series` + `compute_ic_hac_stats` as the replacement. Using every decision minute
and the HAC correction, the trailing 15-minute return earns a mean IC of **-0.00798** with a HAC
t of **-6.02** (19 Bartlett lags) against a naive **-16.78**. The old cell reported a near-zero,
not-significant baseline; the corrected one is small but clearly separated from zero.

This refines rather than reverses the case study's story: the reconciliation with Ch6's much stronger
close-to-close reading still holds - midprice removes most of the apparent reversal - and a real but
tiny ranking edge against a spread that does not shrink with the horizon is the tension the case
study exists to work through.

`N_eff` is now `calculate_label_uniqueness` via `case_studies/utils/label_diagnostics`, replacing a
hand-rolled formula: 14,370,375 primary-label development rows carry 1,062,039 effective
observations (ratio 0.0739 against the 0.0714 that fourteen fully overlapping intervals imply).

What a label consumes is **return intervals, and one fewer than its horizon in bars** - entering a
bar after the decision and leaving at the horizon spans the moves between those two prices, so
`M[t+H]/M[t+1] - 1` telescopes over the intervals `t+2 ... t+H`, which is `H-1` of them. The first
draft of this notebook passed `horizon=H` to `effective_sample_size`, which models the
`P[t+H]/P[t]` convention that consumes `H`; the ratio then landed near `1/H` circularly, because
that is what it had been told to assume. Caught by the pre-push review and corrected here, along
with the reference ratio, the audit record's overlap row and the prose that explains it. This is the
trap the stage document flags as "the `-1` is load-bearing and getting it wrong is not visible in
the aggregate".

## Conformance

```
$ python3 scripts/check_notebook_conformance.py --stage 02 --case-study nasdaq100_microstructure -v
=== stage 02 (02-labels.md): at most 330 code lines, 4+ figures, 11 required sections ===
  PASS  nasdaq100_microstructure
```

From 3 failing checks and 13 findings (0 figures against 4 required, 4 numbers in untagged prose, all
8 A-H sections missing) to zero.

- **Sections A-H**, in order, with the three tagged `results` cells in E, F and G.
- **Five figures** where printed tables stood: label validity at the window boundary (which is where
  defect 1 is visible at a glance - the 5-minute curve used to lie exactly on the 15-minute one),
  the distribution across horizons, move scale against the round-trip spread, class shares through
  time, and overlap decay. F5 is correctly absent: resolution time here is `t + h` by construction.
- **Everything bound from `config/setup.yaml`** - label set, horizons as durations, holdout boundary,
  and the flat band from `costs.friction_floor_bps`. No literal.
- **The holdout seal is on `_label_end`**, per label rather than one boundary for all three, since a
  60-minute window closes three quarters of an hour after the 15-minute one opened from the same bar.
- **Library adoption**: `fixed_time_horizon_labels`, `calculate_label_uniqueness`,
  `cross_sectional_ic_series`, `compute_ic_hac_stats`, `panel_autocorrelation`, and `write_artifact`
  for the digest sidecars each label now carries.
- **Removed**: the dead 38-line `results` dict (the vestige of the banned `results/*.json` format),
  the naive t-statistic, the hand-rolled effective-sample-size formula, and the unused
  `generate_cv_splits` call - nothing reads a `cv_config.json` here and none is written, with
  Section H stating positively where the folds do come from.

`tests/test_holdout_boundary.py` gains this notebook in both static gates, per the stage-02 CI
contract. It stays out of `PER_SYMBOL_ENDPOINT_NOTEBOOKS`, whose regex binds a single quoted entity
column and so cannot express this notebook's compound `["symbol", "session_date"]` key; the comment
there records why.

## Two things this notebook does not settle

The label uses a one-minute execution delay while `config/setup.yaml` declares the decision cadence
as 15-minute bars with a one-bar delay. Read the second way, `fwd_ret_5m` would resolve before the
position could be entered and `fwd_ret_15m` would have a zero-length holding period. The stage-02
standard lists this case study's convention as entry-bar-mid and this rebuild keeps it, because
changing it moves every label value and is a statement about what the case study teaches rather than
a conformance fix. It needs a ruling, and the two documents should then say the same thing in the
same units.

The round trip in Section E is twice the half-spread quoted at the decision bar, not the two spreads
actually crossed at the entry and exit bars. The known-limitations note now says so. It does not
affect that section's claim, which is that the median spread does not change with the horizon while
the move does.

## Figures were reviewed as rendered, not as code

Three title claims were wrong against the data and were corrected after looking at the PNGs: most
moves at every horizon turn out to be *larger* than the spread rather than smaller, the flat class is
the *smallest* of the three rather than dominant, and the 60-minute label's overlap does not decay to
zero at its horizon - it crosses zero near lag 50 and sits at -0.219 by lag 60, which is a
finite-session artifact the notebook now explains rather than a claim it can make.

### DIFF
--- a/case_studies/nasdaq100_microstructure/02_labels.ipynb
+++ b/case_studies/nasdaq100_microstructure/02_labels.ipynb
@@ -51,10 +51,10 @@
    "id": "b6dfca90",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-07-31T09:11:34.929354Z",
-     "iopub.status.busy": "2026-07-31T09:11:34.928932Z",
-     "iopub.status.idle": "2026-07-31T09:11:36.891522Z",
-     "shell.execute_reply": "2026-07-31T09:11:36.891085Z"
+     "iopub.execute_input": "2026-07-31T09:30:02.592837Z",
+     "iopub.status.busy": "2026-07-31T09:30:02.592403Z",
+     "iopub.status.idle": "2026-07-31T09:30:04.540543Z",
+     "shell.execute_reply": "2026-07-31T09:30:04.539818Z"
     },
     "papermill": {
      "duration": 1.580942,
@@ -109,10 +109,10 @@
    "id": "810858da",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-07-31T09:11:36.893209Z",
-     "iopub.status.busy": "2026-07-31T09:11:36.893115Z",
-     "iopub.status.idle": "2026-07-31T09:11:36.894676Z",
-     "shell.execute_reply": "2026-07-31T09:11:36.894368Z"
+     "iopub.execute_input": "2026-07-31T09:30:04.542001Z",
+     "iopub.status.busy": "2026-07-31T09:30:04.541888Z",
+     "iopub.status.idle": "2026-07-31T09:30:04.543725Z",
+     "shell.execute_reply": "2026-07-31T09:30:04.543406Z"
     },
     "papermill": {
      "duration": 0.004977,
@@ -155,10 +155,10 @@
    "id": "1457d387",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-07-31T09:11:36.895641Z",
-     "iopub.status.busy": "2026-07-31T09:11:36.895581Z",
-     "iopub.status.idle": "2026-07-31T09:11:36.905514Z",
-     "shell.execute_reply": "2026-07-31T09:11:36.905231Z"
+     "iopub.execute_input": "2026-07-31T09:30:04.544724Z",
+     "iopub.status.busy": "2026-07-31T09:30:04.544664Z",
+     "iopub.status.idle": "2026-07-31T09:30:04.554747Z",
+     "shell.execute_reply": "2026-07-31T09:30:04.554428Z"
     },
     "papermill": {
      "duration": 0.003947,
@@ -185,10 +185,10 @@
    "id": "d1fea063",
    "metadata": {
     "execution": {
-     "iopub.execute_input": "2026-07-31T09:11:36.906506Z",
-     "iopub.status.busy": "2026-07-31T09:11:36.906448Z",
-     "iopub.status.idle": "2026-07-31T09:11:36.909029Z",
-     "shell.execute_reply": "2026-07-31T09:11:36.908733Z"
+     "iopub.execute_input": "2026-07-31T09:30:04.555781Z",
+     "iopub.status.busy": "2026-07-31T09:30:04.555719Z",
+     "iopub.status.idle": "2026-07-31T09:30:04.558524Z",
+     "shell.execute_reply": "2026-07-31T09:30:04.558200Z"
     }
    },
    "outputs": [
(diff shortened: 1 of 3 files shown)


stefan-jansen:
**Do not merge yet: `cs-nasdaq100_microstructure` fails on the test fixture, and the failure is correct.**

The required checks (`guards`, `lint`, `test-unit`) pass. The case-study job does not:

```
Error: Cell 13 (AssertionError): fwd_ret_5m: a 0:05:00 horizon is 1 bar on a 0:05:00 grid,
so the entry bar and the exit bar are the same bar and the label cannot be formed
```

The `nasdaq100/minute_bars` test fixture is on a **5-minute** grid, not a minute grid: every symbol-session in it holds 78 bars spaced five minutes apart, against 390 one-minute bars in production. That is why this PR trips the assertion, and it is also why the job was green before - the previous code counted rows, so on that fixture `fwd_ret_5m` was silently a 25-minute label and `fwd_ret_15m` a 75-minute one. The gate was passing while exercising labels the case study does not define.

On a 5-minute grid no execution convention with a non-zero delay can express a 5-minute label: the first tradable price after the decision is already the exit. So the fixture cannot represent this case study's shortest label, and the fix is to regenerate that dataset at 1-minute resolution together with the intermediates derived from it.

Weakening the assertion would restore a green that measures nothing, so I have left it failing and raised the fixture regeneration separately. Production is unaffected: the notebook was executed end to end on the full 114-symbol panel, where the grid is exactly uniform and all four labels are formed.

stefan-jansen:
Carried onto the group A integration branch `group/csA-features`, which collapses the four stage-02 label branches into one. This PR's commits are on it as e5190037, d3a42049, with the case study's files byte-identical to this branch.

PR #469, opened 2026-07-31
Unblock the merge queue: the reduction reaches the notebook, the sanitizer spares source, and there is one ruff

stefan-jansen:
`test-unit` is a required check and it has been red on `main` since #428, so **no PR in this repository can merge**. This makes it green, and takes the notebook-hygiene gate off its CI deselect while it is here, since both are the same class of defect and the same job.

## Twelve Papermill parameters never reached their notebook

```
tests/test_pm_helpers.py::test_every_declared_parameter_reaches_its_notebook FAILED
```

CI declares those parameters in order to *reduce* a notebook for test mode. A parameter that is accepted, injected and then discarded means the notebook runs unreduced while the harness believes it set a reduction - which is exactly what `test_every_declared_parameter_reaches_its_notebook` (added in #446) exists to catch. Two mechanical faults, both introduced by #428.

**Four notebooks lost their `# %% tags=["parameters"]` cell.** Compare `75b659b5~1`: `14_latent_factors/04_ipca`, `20_strategy_synthesis/01_aggregate_synthesis`, `06_cost_survival` and `07_regime_risk` each had one. `05_portfolio_allocation`, untouched by that PR and still passing, has the identical shape to 06 and 07. Without the tag papermill injects above the imports and the notebook's own binding overwrites it a few lines later. The tag is restored on the cell that already binds the declared names.

`04_ipca` needed one thing more. It used to derive its split from `TRAIN_FRAC`; #428 replaced that with the literal `TRAIN_BOUNDARY = 350`, so the declared `N_PERIODS: 200` would have started the evaluation window 150 periods past the end of the panel. `TRAIN_BOUNDARY: 140` is declared alongside it, the same 0.7 of the panel production splits at.

**`07_stochastic_discount_factor` kept its cell, and #428 renamed the knobs under it**, leaving `overrides.yaml` declaring five names the notebook no longer has. Renamed to the ones it does, at the same values: `NUM_EPOCHS_UNCONDITIONAL`→`UNCONDITIONAL_EPOCHS`, `NUM_EPOCHS_INSTRUMENTS`→`INSTRUMENT_EPOCHS`, `NUM_EPOCHS_CONDITIONAL`→`CONDITIONAL_EPOCHS`, `NUM_ADVERSARIAL_ROUNDS`→`ADVERSARIAL_ROUNDS`, `BETA_NET_EPOCHS`→`BETA_EPOCHS`. `EVAL_EVERY: 4` joins them: production evaluates every 16 epochs, and at 8/12/8 epochs that is never, so the validation checkpoint the notebook selects would never be reached and CI would exercise none of it.

`N_MACRO_COLS` is dropped rather than renamed. The rewritten notebook fixes its macro state to eight named daily series with no count to set, and `grep -rn N_MACRO_COLS` over the repository returns only the declaration itself.

## The path sanitizer rewrote source, so its test was deselected

`test_no_machine_specific_paths_in_committed_notebooks` has been red and deselected since the count in the workflow comment was 97 leaks across 46 notebooks; it is 108 across 53 today. Readers see `/home/stefan/ml4t/code/...` and Docker `/app/...` in cell outputs and papermill metadata.

The deselect was right at the time and said so: the fix was **not** to run `sanitize_notebook_paths.py`, because it rewrote the raw `.ipynb` text and therefore rewrote source too - it turned the real Docker mount path in `02_financial_data_universe/16_provider_comparison` into a relative path. So the sanitizer is what changes here. It now parses the notebook and considers only strings reachable through notebook metadata, cell metadata and cell outputs. Two strings in the repository's source survive on exactly that basis: that mount path, and the comment in `12_gradient_boosting/02_gbm_comparison` that names `/app/` to explain it.

The rewrite is still applied to the raw text one substring at a time, so **the diff is the replaced paths and nothing else** - no reserialization, no reflowed formatting, no reordered keys. A string the source also contains is skipped and named rather than replaced, since a raw-text edit cannot tell the two occurrences apart; nothing in the repository trips that today, and a unit test pins the behaviour.

**No `.py` changes and no notebook's `source` differs from its committed version**, checked cell by cell across all 53.

## One ruff

`uv run ruff` resolved 0.15.14, the pre-commit hook pinned 0.15.8, and the `lint` job ran `uvx ruff@0.15.8`. Two ruff versions format the same code differently, and because jupytext embeds the `.py` source in the `.ipynb`, the hook's counter-reformat at commit time turns a freshly executed notebook into a stale render of its own source and the notebook-sync gate rejects it - one wasted re-execution cycle per notebook PR. All three are 0.15.14 now, exactly pinned, and `tests/test_toolchain_pins.py` fails if any drift or if the lock stops resolving the pin.

## Not in this PR

`/tmp` paths in 62 notebooks' outputs are a different rule from the one this gate enforces, and only two of their four categories are leaks - `/tmp/ml4t-test-output` is the documented `ML4T_OUTPUT_DIR`, and rewriting it would repeat the mistake this PR is fixing. Filed separately with the per-category evidence.

## Verified

```
uv run pytest tests/test_pm_helpers.py tests/test_toolchain_pins.py \
              tests/test_notebook_output_hygiene.py -q     # 81 passed
grep -c deselect .github/workflows/test.yml                # 0
python .github/scripts/sanitize_notebook_paths.py --check
# clean: no machine-specific paths in any notebook's outputs or metadata
uvx ruff@0.15.14 check utils/ data/ tests/                 # All checks passed!
pre-commit run ruff-format --all-files                     # Passed
```

Plus the full `test-unit` set locally: 236 passed. `04_ipca.ipynb` is re-executed because the provenance gate reads the tag as a source change; every text output is byte-identical to the committed one.

### DIFF
--- a/.github/scripts/sanitize_notebook_paths.py
+++ b/.github/scripts/sanitize_notebook_paths.py
@@ -8,8 +8,26 @@ rewrites them in place:
 * repo-internal paths -> repo-relative (matching ``utils.paths.display_path``)
 * anything else under ``~/ml4t`` -> a ``~``-prefixed generic path
 
-It edits the raw ``.ipynb`` text (no JSON reserialization) so the only diff is
-the replaced substrings — formatting, key order and outputs are untouched.
+Outputs and metadata only; ``source`` is never touched
+------------------------------------------------------
+
+A notebook's ``source`` is code the reader runs, and a path in it can be load
+bearing. An earlier version of this script rewrote the raw ``.ipynb`` text and so
+rewrote source along with everything else: it turned the real Docker mount path
+in ``02_financial_data_universe/16_provider_comparison`` into a relative path.
+That is why the companion test was deselected in CI rather than fixed.
+
+So the notebook is parsed, and only strings reachable through notebook metadata,
+cell metadata and cell outputs are candidates. Two strings in this repository's
+source survive precisely because of that: the mount path above, and a comment in
+``12_gradient_boosting/02_gbm_comparison`` that names ``/app/`` to explain it.
+
+The rewrite itself is still done on the raw text, one substring at a time, so the
+diff is the replaced paths and nothing else - no JSON reserialization, no
+reflowed formatting, no reordered keys. A string that also occurs inside the
+notebook's source is skipped and reported instead of replaced, because a raw-text
+edit cannot tell the two occurrences apart. Nothing in the repository trips that
+today; the check is here so the guarantee holds without being re-verified by hand.
 
 Idempotent: running twice is a no-op. A companion test
 (``tests/test_notebook_output_hygiene.py``) fails CI if any leak survives.
@@ -22,6 +40,7 @@ Usage:
 from __future__ import annotations
 
 import argparse
+import json
 import re
 import sys
 from pathlib import Path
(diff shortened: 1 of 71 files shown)


stefan-jansen:
Merging on the three required checks - `test-unit`, `guards` and `lint` - all green.

The notebook matrix is not required and was already red on `main` at the #428 merge ([run 30650971953](https://github.com/stefan-jansen/machine-learning-for-trading/actions/runs/30650971953)). This branch improves it: `ch18-20` goes from ten failing notebooks to five, because restoring the parameters cell lets `ALLOW_MISSING_SPINE: true` reach `01_aggregate_synthesis` again, which also unblocks `02_feature_evaluation`, `04_signal_to_strategy` and `05_portfolio_allocation`. `ch14-15` is unchanged at one, in a notebook this PR does not touch.

`14_latent_factors/04_ipca` passes under the reduction that had never been applied before, which is the check that `TRAIN_BOUNDARY: 140` was both needed and right.

The six that remain are filed with their cell numbers and failure messages, split by likely cause; two of them are the CI ETF fixture holding seven symbols where the notebook asks for `VTV`/`VUG` and `VNQ`.

PR #470, opened 2026-07-31
Ch2-3: figure titles carry the claim, and the house palette actually applies

stefan-jansen:
Figure-quality pass over Chapters 2 and 3 (the `ch02-03` CI job), 20 notebooks re-executed on
production data. No computed result changes anywhere in this branch.

## What was wrong

**Figure titles carried values the notebook's own code had just produced** — a correlation, a PSI,
an observation count, a Monte Carlo draw count, a traded-value share. Those read as results the
reader has to re-verify against the chart, and the house rule puts the claim in the title and the
numbers in the axes. 53 titles across the two chapters; six also ran over the 75-character
ceiling, up to 97.

**Four notebooks never imported `utils.style`**, which is what registers the `ml4t` Plotly template
and makes it the default. Every figure in them rendered on the stock template, and four then
pinned `plotly_white` on top of that. They also carried raw hex and CSS color names — `#3498db`,
`#e74c3c`, the Bitcoin/Ethereum/Solana/Binance brand colors, `red`, `gray`, `darkblue` — instead of
the locked palette.

## What changed

- Titles now state a claim, or stay descriptive where the chart does not support one.
- `utils.style` imported where it was missing; raw colors mapped onto `COLORS`.
- `16_provider_comparison` plots seven ETFs against a six-color colorway, so SPY, GLD and IWM
  shared one navy. Identity now comes from a direct end-label and the legend is gone — the palette
  is deliberately capped, so the fix is not a seventh hex.
- `11_crypto_premium_analysis`: SOL's premium runs to a few thousand basis points and `nbinsx`
  takes its bin width from the full range, so that panel was two bars wide. Dropping the outer 1%
  **before** binning is what makes all four distributions legible; clipping the axis alone looked
  correct in the source and was still wrong in the render.
- `14_itch_bar_sampling` called `relative_to(ML4T_PATH)` on the data directory, which raises
  whenever `ML4T_DATA_PATH` points outside the repo — the standard multi-worktree setup, and any
  reader who keeps data on another volume. It now falls back to the absolute path.

## On the titles that got weaker, not stronger

Three titles were reverted to descriptive or had a qualifier restored, because the rendered chart
did not support what the words said:

- the OFI scatter in `03_itch_lob_analysis` has a near-flat trend at ρ=0.12, so "move together" was
  true of the statistic and false of the picture; same for the TAQ imbalance panel at r=0.08;
- `14_itch_bar_sampling` claims activity-sampled bars sit closer to normal while the code truncates
  every return to ±2%, and tail truncation is the whole subject of a normality claim;
- `17_databento_bar_sampling` plots `.head(100)` per bar type, giving each series a different end
  time, which without the qualifier reads as a like-for-like comparison.

Every title was checked against the extracted PNG rather than the source. That is where all three
came from, and one of them only after review pushed back.

## Verification

Chapter 2 §2.3's survivorship-bias estimate was re-derived independently from the parquet, in
pandas, from the notebook's stated method rather than its code, before anything was touched:
raw −13.6 pp, repaired +8.3 pp, and +8.0 / +3.5 / +15.3 pp across the three scenarios — matching
the notebook exactly, including 805 untrusted returns of 2,931,079. The sign also survives two
repair rules the notebook does not use, so it is the repair that sets it rather than the specific
threshold. The script is committed in the agent workspace.

The four dataset-count claims that had never been checked against their parquet all hold: 100
ETFs, 20 FX pairs / 478,640 rows, 19 perps / 866,484 rows, 30 CME products.

All 20 re-executed notebooks: 0 errors, 0 unexecuted cells, figure counts equal to their
pre-change baseline, every Plotly figure carrying its PNG sibling. Provenance stamped on each.


stefan-jansen:
## CI status

**Every Chapter 2 and Chapter 3 notebook passes.** The `ch02-03` job is red on one notebook that
is neither in this diff nor in these chapters:

```
FAILED tests/test_chapter_notebooks.py::test_chapter_notebook[18_transaction_costs::03_market_impact_calibration.py]
Error: Cell 42 (ValueError): Too few positive normalized coefficients for cross-sectional inference.
```

That job's matrix reaches past its name. The same test fails on `main` — run
[30664800602](https://github.com/stefan-jansen/machine-learning-for-trading/actions/runs/30664800602),
where it is the only failure in the job — so `ch02-03` cannot go green for any PR until it is
fixed. Tracked separately.

### One regression of mine, found by CI and fixed

The first run also failed `02_financial_data_universe/11_crypto_premium_analysis` with
`IndexError: index -1 is out of bounds for axis 0 with size 0`. Mine: the percentile clip I added
for the SOL panel runs on a reduced fixture where a major symbol can have zero rows, and the cell
handled that case before my change. The clip now applies only when there is data, which restores
the previous behavior for an absent symbol and changes nothing on the production panel. Fixed in
`cc4af3a2`; that test passes.

### Gate rounds

Four, and three carried real findings, all fixed rather than waived:

1. a title asserting "the last four years" against a chart with five bars;
2. two titles that dropped material sampling qualifiers — a ±2% tail truncation under a normality
   claim, and `.head(100)` under a cross-bar-type comparison;
3. output paths still naming a specific checkout, plus the point that mattered — the paired sources
   printed raw `Path` objects, so the sanitizer was hiding a leak that every re-run recreated. Those
   seven notebooks now use `utils.paths.display_path`.

The final push passed the gate on its own.

PR #472, opened 2026-08-02
Bring the Chapter 6-7 teaching notebooks to the standard

stefan-jansen:
Six notebooks in Chapters 6 and 7, reviewed as one group against the notebook
standards, corrected, and re-executed on production data. One PR, because the
chapters are independent and the CI notebook matrix already bundles them as
`ch06-07`.

Every notebook here executes clean end to end (`errors=0 unexecuted=0`) and every
number quoted below was recomputed from the committed run, not carried over.

## What was wrong

The defects fall into one class: **the notebook demonstrated something other than
what it said it demonstrated.** Not crashes - each of these ran green and produced
plausible output.

### `06/02_cv_foundations`

- The notebook whose thesis is "count in trading days" reported its purge as an
  index difference. Prose said 21 NYSE trading days; the cell four screens later
  printed `gap = 22 trading days` from the same `label_horizon=21`. Purging 21
  sessions leaves 21 rows between the endpoints, so the endpoints sit 22 apart. It
  now prints the sessions withheld, and the markdown names the off-by-one instead
  of leaving a reader to resolve it by picking one.
- The CPCV figure plotted `np.random.randn` on its y-axis, with split membership
  carried only by colour - inviting the reader to look for structure that was
  randomly generated. A split is a partition and has no value to plot. Replaced by
  a 15-split block-occupancy heatmap.
- "15 splits produce 5 backtest paths" was the section's most-repeated claim and no
  cell ever showed a path. A round-robin factorization now builds and prints them.

### `07/01_data_quality_diagnostics`

- **A check that could not fail.** `check_index_integrity` sorted by time and then
  asked whether time was sorted. On a deliberately non-monotonic frame
  (`{'s':['A']*3,'t':[3,1,2]}`) it returned `True`; without the pre-sort, `False`.
  Six `Index Integrity: PASSED` lines rested partly on it, in a notebook whose first
  takeaway is that monotonicity must pass.
- **A verdict that was a literal.** The firm-characteristics index check was
  `{"passed": True, "note": "No symbol column (cross-sectional dataset)"}`. The
  dataset has a `symbol` column. Running the real check: 9,861 firms, 804,530
  firm-months, all four sub-checks true - so the verdict was right by luck and its
  stated reason was false.
- **A takeaway inverted by the data.** "Extreme returns often signal corporate
  actions", behind a `|return| > 1.0` filter. A raw return is bounded below by −1,
  so that filter can only ever fire on the upside. Measured on the production panel:
  837 rows flagged, smallest `+1.0001`, of which 382 are split days; but there are
  3,644 forward splits with a median same-day return of −0.4850, and the filter
  catches **1** of them. `split_ratio` was sitting unused in the same DataFrame.
- Heatmap captions described a "staircase", a "diagonal" and "vertical bands …
  exchange closures". Symbols were in alphabetical order so there was no staircase
  and no diagonal, and the axes are x=asset/y=month, so a vertical band would be an
  asset rather than a date. Ordered by first observation, the staircase is real and
  the captions now describe it - including the 20 interior gaps, which a first draft
  of the replacement caption wrongly denied.
- **And the cell added to make those captions checkable could not fail either.** It
  counted "months missing for every asset" off a pivot that emits one row per month
  that *has* data - so a month in which the whole panel is missing, which is exactly
  what the caption invokes, is absent from the matrix rather than present as an
  all-zero row. The count was zero by construction, in the notebook whose subject is
  checks that cannot fail. The matrix is now reindexed onto a continuous month range
  before anything reads it. Discriminator, on a frame with one month deliberately
  absent: `0` before (unfalsifiable), `1` after. The companion "assets missing in
  every month" line is removed rather than fixed - the symbol axis is the observed
  universe, so the quantity is not expressible here, and a guaranteed zero beside a
  real count is worse than nothing.
- Deleted 72 lines of `calendar_gap_check` / `_find_gaps`, fully documented under
  their own section and never called - carrying `if weekday == 4:  # Friday`, the
  pandas Mon=0 convention, i.e. the exact off-by-one that caused the FX weekend
  defect, waiting in dead code to be revived.

### `07/04_maximum_favorable_adverse_excursion`

- **Excursions measured across three interleaved contracts.**
  `load_cme_futures(products=["ES"])` returns one row per (session, tenor): 11,458
  rows over 3,866 sessions, tenors 0/1/2. The notebook sorted by timestamp with no
  tenor filter, and `compute_mfe_mae` walks its forward window with `shift(-k)` over
  whatever row order it receives - so the 21-bar window stepped across the front,
  second and third contract rather than forward in time on one instrument, and the
  gaps between contracts registered as favorable and adverse moves. Filtering to the
  front contract before the sort, and with the window correction below: MFE p90
  18.49% → **6.56%**, MAE p75 13.54% → **4.43%**, MAE p90 15.94% → **7.94%**. These reach the chapter through
  `mfe_mae_summary.json`, which the notebook designates as the source for chapter
  references, so the barrier-width references for ES were roughly three times too
  wide.
- **Barrier configs were round numbers wearing percentile labels.** The validation
  ran `("Tight (p50)", 0.02, 0.01)`, `("Medium (p75)", 0.04, 0.02)`,
  `("Wide (p90)", 0.06, 0.03)` under the comment *"Test different barrier widths
  based on percentile analysis"*, while the percentiles were computed forty cells
  earlier and never referenced. The measured widths are 3.05/1.98, 4.64/4.21,
  **6.51/7.80** - note the last pair: at p90 the stop belongs *wider* than the
  take-profit, because SPY's adverse excursions have the heavier tail. The hardcoded
  `0.06/0.03` asserted the opposite ordering, so the round numbers did not merely
  approximate, they inverted the relationship the section exists to establish.
- **A recommendation with nothing behind it.** Section 9 recommended `ES: 1.5xATR`
  take-profit and `1xATR` stop, rationale "tighter for institutional flow". ATR was
  computed once in the whole notebook, on SPY. With ES ATR now computed at 1.39%,
  the stop the measurement supports is **3.44x** ATR, not 1x - and ES adverse
  excursions turn out to be *wider* than SPY's at both the median (2.16% vs 1.98%)
  and p75 (4.43% vs 4.21%), so a tighter stop for ES inverts the measurement. The
  same defect as the take-profit/stop ordering above, one section later. The table
  is now built from each instrument's own quantiles in a code cell.
- **The excursion window was 22 bars, and it opened at the entry bar.**
  `compute_mfe_mae` built its forward extremes from `shift(-k) for k in
  range(horizon_bars + 1)` - shifts 0 through 21. The entry price is `close(t)`, so
  bar t's own high and low are *pre-entry*: they measure movement the position was
  never exposed to, and they stretch a stated 21-bar holding period over 22 bars.
  The docstring and the header equations both defined the window as `[t, t+H]`, so
  the code matched its documentation and the documentation was wrong.

  The correction has a signature worth pointing at: the high quantiles are
  **unchanged to sixteen decimal places** (SPY MFE p75 4.6360580506486615, p90
  6.513815723616401), because a large excursion is almost never driven by the entry
  bar, while the low quantiles move sharply - SPY MAE p25 0.935% → 0.735%. BTC moves
  most, which is the expected direction: at an 8-bar horizon the entry bar is one
  ninth of the window rather than one twenty-second.
- **The ATR multiple was a ratio of aggregates.** The table divided the p75 MAE by
  the mean ATR. A stop placed at `k x ATR(t)` is re-sized every day, so what has to
  hold at the 75th percentile is the ratio `MAE(t)/ATR(t)`, not the ratio of two
  separately aggregated numbers - and the two differ because ATR and the excursion
  it scales move together. Replaced with the p75 of the per-entry ratio.

### `07/06_ic_inference`

`compute_ic_hac_stats(ic_series)` was called without `label_horizon`, in the
notebook whose subject is HAC adjustment. The library's automatic Newey-West rule
`floor(4*(T/100)^(2/9))` reads only sample size and returns L = 9 at T = 4,989 - a
rule that never looked at the labels or the signal. The labels are 21-day
*overlapping* forward returns and the signal is 21-day momentum, whose
cross-sectional ranks change slowly; overlapping outcomes measured against
persistent ranks give ICs that move together over roughly the label horizon. The
library already supports `label_horizon` and warns when it is omitted.

The notebook is careful that **both halves of that sentence are load-bearing**. The
IC is a rank correlation, not a return: if the signal ranks were redrawn
independently each day, the ICs could be serially uncorrelated no matter how much
the return windows overlap. So `h-1` is presented as a conservative horizon-aware
choice for a persistent signal, justified by the measured ACF (outside the
white-noise band to about lag 16), rather than as a theorem about overlapping
labels.

```
   bare (auto lag): lags=  9  hac_se=0.011218  t=0.0732  p=0.9416
  label_horizon=21: lags= 20  hac_se=0.013230  t=0.0621  p=0.9505
```

The notebook printed `HAC SE 0.0112` beside `Bootstrap SE 0.0132` and attributed the
gap to estimator family. The bootstrap was right. The two CIs now agree
([-0.0251, 0.0268] vs [-0.0253, 0.0271]), which is what the figure title already
asserted. `21` had also been retyped in four places; one `LABEL_HORIZON` now drives
all of them plus the HAC truncation.

### `07/07_multiple_testing`

Three defects, all of the "demonstrates the opposite" class.

- **RAS mixed two scales inside one subtraction.** The Rademacher complexity was
  computed from a standardized IC matrix and subtracted from raw ICs whose std is
  0.0075, giving a penalty of 0.3143 against a best observed IC of 0.0198 - a 16x
  deduction that rejects every candidate whatever the data says. The Massart-bound
  ratio *does* need the standardized scale, so the fix is two complexities: 0.0453
  in IC units for the deduction, standardized for the ratio.
- **The DSR table put an annualized Sharpe next to a per-period one.** `Best Sharpe
  (ann.) 1.30` printed directly above `E[max] under null 0.08`. The library returns
  `expected_max_sharpe` per period: 0.0774 × √252 = 1.23. The table invited a
  comparison off by a factor of 15.9, and made a near-total selection effect look
  like a small one.
- **The MinTRL demo disabled the adjustment it exists to demonstrate.**
  `variance_trials=0.0` collapses the FWER correction to the single-test case, so the
  N=1 / N=10 / N=100 columns were identical to the last decimal in a section headed
  "additionally adjusts for the number of strategies tested". Sourced from the actual
  dispersion of the 50 strategies searched, N=10 at SR=1.0 goes 2.7y → 121.0y and
  N=100 → never.

- **`kappa` bounded the wrong quantity.** The RAS estimation term needs a bound on
  the *per-period* IC observations that get averaged, not on the averaged IC.
  `kappa=0.05` was chosen because it "bounds the largest observed |IC|" - true of the
  averaged ICs (max 0.022) and irrelevant; a per-period Spearman IC is supported on
  [-1, 1] and the observed maximum here is 0.5455. This is the same units confusion
  as the complexity above, one term to the right in the same expression. The reported
  inference now uses `kappa=1.0`, which holds by construction: a bound read off the
  sample is a function of the sample, so it carries no coverage guarantee at all, and
  the significance flag downstream would mean nothing. The empirical value is shown
  beside it as a labelled sensitivity row, because the gap between the two is the
  honest cost of a distribution-free bound on one year of data.
- **The stated standardization scale was the wrong statistic.** The prose said
  standardizing inflates the complexity "by roughly the IC standard deviation,
  around 0.008". The divisor is each factor's IC dispersion *across dates* (0.143);
  0.008 is the dispersion of the *averaged* ICs *across factors*. Both are printed
  now, alongside the implied ratio R_raw/R_std = 0.144, which agrees with the divisor
  to three decimals.

No verdict flips: RAS still rejects all 100 noise factors, DSR still returns
not-significant, BH-FDR still finds 15 of 15 true factors. Every moved number was
being displayed on the wrong scale, with the correction switched off, or against a
bound that was not one.

### `07/08_causal_sanity_checks`

- **The permutation p-value was printed finer than the procedure can resolve.**
  `np.mean(|null| >= |observed|)` over B=200 printed `p=0.000` and the prose read
  "p < 0.001". With B=200 the finest resolvable non-zero value is 0.005. Now
  `(r+1)/(B+1)` at B=1000, with the floor printed next to the value; momentum sits
  *at* the floor, which is the strongest statement the procedure can make and is now
  labelled as such.
- **The null ignored the label's temporal dependence.** Shuffling assets within each
  date implies the per-date ICs are independent and shrinks the null's spread like
  σ/√n, but the labels are 21-day forward returns sampled daily. Replaced with a
  block permutation at the label horizon, matching the precedent already in
  `05_signal_evaluation`.
- **Thirty tests, no multiple-testing correction** - in the notebook immediately
  after `07_multiple_testing`, whose lesson its own opening cell cites. The heatmap
  starred cells at bare `|t| > 2`. Under BH-FDR: 6 significant at raw p < 0.05, 3
  after correction, against 1.5 expected false positives if all 30 were null. The
  correction removes both 21-day cells the rest of the chapter relied on, so section
  4's findings, the feature-selection rationale and the opening summary were all
  rewritten against the corrected scan.
- **The "1d Reversal, t = 3.0" hit does not survive a shift of its label.** `rev_1d`
  and `fwd_5d` both contain −log P_t, so noise in the day-t close enters feature and
  label with the same sign - a mechanism that induces covariance whether or not any
  reversal exists. Measuring the same five-day holding period one session later takes
  HAC t from **+2.98 to +1.61** and p from 0.0029 to 0.1076.

  The section is careful about what that does *not* show. The shift is a change of
  window, not a removal of the shared price alone: it drops the t→t+1 session **and
  adds a t+5→t+6 session**. At least four things could produce the decline -
  endpoint noise, a real effect in the dropped session, offsetting returns in the
  added one, or sampling variation on a t-statistic that was never large. One
  comparison making several simultaneous changes cannot attribute its own result, so
  the notebook claims only that the cell fails the check, names the measurements that
  would separate the four, and leaves them to Chapter 16 where the execution
  assumptions live. That is enough for the decision it needs to make: a result that
  depends on which five sessions are used is not one to build on.
- **The shared-driver verdict gated on a signal-existence test**
  (`if tsy_pass and perm_pass`), so a near-null feature was flagged for a confound
  when its actual problem is having nothing to confound. The verdict now keys on the
  Treasury arm alone.
- The scan heatmap rendered its y-axis labels clipped - "21d Momentum" as
  "1d Momentum", "63d Momentum" as "3d Momentum" - so a reader would misread which
  feature each row is.

## Three things worth calling out

**1. A false figure title shipped, and only looking at the rendering caught it.**

Replacing the label "Feature × Horizon IC Scan (HAC t-statistics)" on the causal
notebook's heatmap, the worker wrote *"Only long-lookback momentum survives
correction, and only at 5 days"* - a claim, the right length, correctly formatted,
and false. Three cells carry a BH star and the third is `1d Reversal` at t = 3.0, not
momentum. The title had been written from the conclusion expected rather than from
the grid, and it took opening the PNG to see the third star. It is now *"Three of 30
cells survive FDR correction, all at the 5-day horizon"*, which is countable off the
chart.

The notebook's own prose named all three survivors correctly in the same edit. The
error was in the one artifact no checker reads. That is the defect class this group
kept finding - the CPCV noise y-axis, the DSR scale mismatch, the MinTRL identical
columns, the clipped heatmap labels are all invisible to a linter and visible in ten
seconds of looking.

Since it is that class, this PR also takes the hand-maintained counts out of three
other titles that were interpolating computed values and wrapping across the plot
(82, 79 and 96 characters). Each claim is unchanged and still countable off its
chart; what is gone is a number that has to be maintained by hand.

**2. `07/04` is closed in this PR rather than deferred.**

It was the one notebook the lane report signed off **NO**: its two correctness
defects were fixed and verified, but it had not had the rendered-figure pass the
other five received, and the `ES: 1.5xATR` recommendation was unverified. Both are
closed here rather than filed, because it is one notebook in this group and the group
is the unit.

The figure pass found a real defect. Both histogram pairs drew their two panels on
independent x and y ranges while their titles compared the panels, so the narrower
distribution was rendered exactly as wide as the broader one - a title asserting one
runs wider than the other was unverifiable from the figure making the assertion. Both
now share a scale and carry the reference line their claim is about. The BTC
reference line sat at a flat 2.0% commented "from typical crypto settings", a round
number asserted as typical inside the notebook arguing that barrier widths must be
measured; it is now each side's own 75th percentile, 1.71% and 1.74%.

**3. `07/08` went from about 25 minutes to about 3, and the reason is a correctness
fix rather than an optimization.**

Raising the permutation count from 200 to 1000 - which the p-value resolution
required - had pushed the notebook past 26 minutes. Part of that was paid back by
replacing the per-date `np.corrcoef` with a dot product of pre-standardized rank
vectors, identical to 8e-18 and 17x faster, which left it around 25 minutes.

The rest turned out to be a second, unfixed copy of the same statistic.
`_figure_7_10_permutation` - the path that writes `figure_7_10_inputs.npz`, which
the book's Figure 7.10 reads - still ran an independent within-date shuffle scored
with `np.mean(|null| >= |observed|)`. Both of those are the defects the correction
above is about: the iid null is far too narrow against overlapping labels, and the
plain proportion can return exactly `0.000`, which no permutation test can resolve.
So the notebook was reporting a dependence-aware p-value in its own text and an
uncorrected one in the artifact the book prints.

Pointing it at the same block-permutation helper fixes the artifact and, because
that helper is the vectorized one, takes the notebook from **1482s to 190s**. Both
paths now report `p=0.0010` for momentum and `p=0.3247` for reversal, and the
resolution floor is written into the artifact beside the value so a caption cannot
quote a floor as a measurement.

An earlier read of this branch attributed the runtime to `_figure_7_10_timing`. That
was wrong - the timing cell is not the bottleneck, and no runtime item is left open.

## Also in this PR

- **Papermill test-tier reductions restored** for `07_multiple_testing`.
  `tests/overrides.yaml` declared one parameter (`N_RAD_ETF`) against a notebook that
  exposes 14; the other 13 reductions were declared under names the notebook does not
  have, so they never reached it and CI ran a notebook it believed was reduced.
  `unusable_parameters()` now returns `{}` for all 14.
- **Machine-specific paths removed from all six notebooks' outputs.** 13 occurrences.
  `test_no_machine_specific_paths_in_committed_notebooks` was deselected in CI when
  this branch was cut and is enforced on `main` as of #469, so the branch would have
  landed red without this. Applied with `.github/scripts/sanitize_notebook_paths.py`,
  which touches outputs and metadata only; `--check` is clean.

## Verification

```
# guards
python .github/scripts/check_windows_safe_paths.py     # OK
python .github/scripts/check_notebook_pair_sync.py     # OK
python .github/scripts/check_import_resolution.py      # OK: 776 files
pytest tests/test_notebook_output_hygiene.py -q        # 7 passed

# lint
uvx ruff@0.15.14 check utils/ data/ tests/             # All checks passed!
uvx ruff@0.15.14 format --check utils/ data/ tests/    # 184 files already formatted

# test-unit surface
pytest tests/test_download_helpers.py tests/test_firm_characteristics_extract.py \
       tests/test_download_all_dispatch.py tests/test_download_scripts_registry.py \
       tests/test_compose_mounts.py tests/test_download_coverage.py \
       tests/test_plotly_renderer.py tests/test_ch21_rl_environments.py \
       tests/test_import_resolution.py -q               # 133 passed

pytest tests/test_pm_helpers.py tests/test_notebook_output_hygiene.py \
       tests/test_toolchain_pins.py -q                  # 82 passed
python .github/scripts/sanitize_notebook_paths.py --check
# clean: no machine-specific paths in any notebook's outputs or metadata
```

All six notebooks were re-executed on production data via papermill with
`cwd=repo_root` and `OMP_NUM_THREADS=1`, and each carries a provenance stamp; the
`notebook provenance sync gate` passes with no STALE entries. No hook was bypassed on
any commit.

`test_pm_helpers.py` was red on `main` when this branch was cut, for five notebooks in
Chapters 14 and 20. #469 fixed it; it is green here.

The `ch06-07` notebook matrix job is not a required check on this repository -
`guards`, `lint` and `test-unit` are, and all three pass locally against this branch.

## A note on how this was reviewed

Every notebook was re-read after each round of corrections, not just the diff. That
turned out to matter more than the initial review: a large share of what was fixed
here is not a defect in the original code but a **sentence describing a computation
that had since changed** - a caption for a figure that was rebuilt, a rationale for a
constant that was replaced, a cross-reference to a figure that moved.

Three of those were second-order, where a correction's own explanation was wrong. The
most instructive is Section 4.1 of `08_causal_sanity_checks`, which took four passes.
Each pass kept a causal claim and qualified it more carefully; the fix was to delete
it. The comparison shifts a five-day return window, so it changes three things at
once, and no single comparison making three changes can attribute its own result. The
notebook's decision only ever needed "this cell fails the check".

The generalisable habit, if it is useful elsewhere: **after changing a computation,
grep the notebook for the quantity it produces and re-read every sentence that names
it.** And prefer deleting a claim to replacing it with a more careful one - a claim
the notebook cannot check is a liability whether or not it currently happens to be
true.


stefan-jansen:
## The `ch06-07` notebook job, and why it stays red

The three required checks - `guards`, `lint`, `test-unit` - pass. The `ch06-07` notebook job
does not, and `504214c8` changes what it fails on rather than making it pass. Reasoning, since
that is the less obvious choice:

The job failed with `ColumnNotFoundError: "p_hac" not found`, twenty cells after the actual
problem. `08_causal_sanity_checks` builds its shared-driver control from IEF alone and attaches
it with a join that drops rows where it is null. **The CI ETF fixture has 24 symbols and IEF is
not among them**, so the control is empty, the join empties the analysis frame to 0 rows, every
feature x horizon pair falls through the `len(ic_series) < 50` branch, and that branch was the
one row shape that omitted `p_hac`.

The empty frame is the older defect and the worse one. On `main` the scan rows carry no `p_hac`
at all, so nothing read the missing column and **this notebook has been running to completion on
zero rows and reporting success**. This branch's Benjamini-Hochberg correction reads `p_hac`,
which is the only reason the condition became visible.

So the notebook now raises where the control is built:

```
Error: Cell 9 (ValueError): No IEF rows between 2022-01-01 and 2024-01-01, so the
shared-driver control cannot be built and the analysis frame would be empty.
Universe contains: ['AGG', 'DBC', 'EEM', ...]
```

Making the job green instead would mean tolerating an empty frame, which is a test that computes
nothing - the state we were already in. The fixture gap lives in the test-data repo and is filed
there with its reproduction; `ch06-07` is not a required check, so this does not block the PR.

Production is unaffected. The canonical universe carries IEF with 5,031 rows over the full range,
so the guard cannot fire; the notebook was re-run on production data (papermill, `cwd=repo_root`,
canonical `ML4T_DATA_PATH`, `OMP_NUM_THREADS=1`), `errors=0 unexecuted=0`, re-stamped and
sanitized, and **the diff carries no result number**.

The other ten red jobs on this run are pre-existing on `main` and unrelated to these chapters -
`ch02-03` fails in `18_transaction_costs/03_market_impact_calibration`, which this branch does
not touch.

stefan-jansen:
## `ch06-07` is fixed in the fixture, not in the notebook

Following the project standard: *"The reason a reduced CI run breaks a notebook is almost never
the notebook. It is the reduction, and the reduction is ours... the fixture data is a repo we own.
`The fixture cannot support this` is a statement about a file we can edit."*

Crossing into the test-data repo was authorized for this item. Fixed there as `e5d72a8`
(`ml4t/third-edition-test-data@main`): the ETF fixture now carries **IEF**, the sole input to this
notebook's shared-driver control, whose absence was emptying the analysis frame to 0 rows.

**Adding IEF alone was not enough**, and this was the trap worth naming. IEF is itself excluded
from the analysis universe as a Treasury, so it restores the control without adding any
cross-section. The scan needs `min_obs=20` symbols per date *and* 50 surviving dates per pair,
against a 21-symbol non-Treasury cross-section with roughly half the CI window burned as warmup by
the 252-day features. So the fixture also gained fifteen long-history non-Treasury names.

40 symbols, 200,634 rows, same 2006-01-03..2025-12-31 window. Every added symbol spans the whole
window, so none introduces a ragged edge; the original 24 are untouched.

**Verified that the scan returns a result, not merely that the error stopped:**

    non-Treasury cross-section   36   (min_obs=20)
    mom_21d  x fwd_5d    459 dates    (threshold 50)
    mom_252d x fwd_5d    228 dates    <- worst case, longest warmup
    mom_21d  x fwd_63d   417 dates

    ML4T_DATA_PATH=~/ml4t/test-data/data uv run pytest tests/test_chapter_notebooks.py \
      -q -k "06_strategy_definition or 07_defining_the_learning_task"
    12 passed, 275 deselected in 168.48s

**Lowering the notebook's thresholds under test parameters was the one option not taken.** It
would have made CI assert something no reader ever runs, and a fixture that agrees with itself
measures nothing.

The guard in `504214c8` stays. It is not the fix, but it is what turns this class of fixture gap
into a failure that names its own cause rather than a `ColumnNotFoundError` twenty cells
downstream - which is how this went unnoticed long enough to ship.

### Retraction, restated

My earlier comment said the other red jobs were "pre-existing on `main`". **That is unsupported
and I withdraw it.** The notebook matrix is skipped on `main`, so it only ever runs on PRs and
there is no green baseline to compare against. What is actually verified is narrower: `ch02-03`
fails inside `18_transaction_costs/03_market_impact_calibration`, a path this branch does not
touch.

Failed jobs have been re-run against the new fixture.

Issue #290, opened 2023-05-07
ML4T Workflow - Vectorized versus event driven - script data.py not found on githup package


gmolinaro-git:
I cannot find the data.py script mentioned in the section "Vectorized versus event driven backtesting". Can someone provide it ?

stefan-jansen:
You are probably looking for https://github.com/stefan-jansen/machine-learning-for-trading/blob/main/08_ml4t_workflow/00_data/data_prep.py

Issue #315, opened 2024-01-22
KeyError: 'No object named quandl/wiki/prices in the file'


frederikplet:
File [c:\Users\frede\Downloads\trading_env.py:75](file:///C:/Users/frede/Downloads/trading_env.py:75), in DataSource.load_data(self)
     [73](file:///C:/Users/frede/Downloads/trading_env.py:73) idx = pd.IndexSlice
     [74](file:///C:/Users/frede/Downloads/trading_env.py:74) with pd.HDFStore('../data/assets.h5') as store:
---> [75](file:///C:/Users/frede/Downloads/trading_env.py:75)     df = (store['quandl/wiki/prices']
     [76](file:///C:/Users/frede/Downloads/trading_env.py:76)           .loc[idx[:, self.ticker],
     [77](file:///C:/Users/frede/Downloads/trading_env.py:77)                ['adj_close', 'adj_volume', 'adj_low', 'adj_high']]
     [78](file:///C:/Users/frede/Downloads/trading_env.py:78)           .dropna()
     [79](file:///C:/Users/frede/Downloads/trading_env.py:79)           .sort_index())
     [80](file:///C:/Users/frede/Downloads/trading_env.py:80) df.columns = ['close', 'volume', 'low', 'high']
     [81](file:///C:/Users/frede/Downloads/trading_env.py:81) log.info('got data for {}...'.format(self.ticker))

File [f:\miniconda3\envs\ml4t\lib\site-packages\pandas\io\pytables.py:596](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:596), in HDFStore.__getitem__(self, key)
    [595](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:595) def __getitem__(self, key: str):
--> [596](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:596)     return self.get(key)

File [f:\miniconda3\envs\ml4t\lib\site-packages\pandas\io\pytables.py:790](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:790), in HDFStore.get(self, key)
    [788](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:788) group = self.get_node(key)
    [789](file:///F:/miniconda3/envs/ml4t/lib/site-packages/pandas/io/pytables.py:789) if group is None:
[...]

PR #480, opened 2026-08-06
cs-etfs: let 05_evaluation reach its own IC floor (1 of 5; job not yet green)

stefan-jansen:
# cs-etfs

**Status: not green.** 1 of 5 real failures fixed. Baseline measured on a quiet box
(load 3-12), 23 notebooks selected: **4 failed, 15 passed, 4 skipped**, 11m02s.

## Per notebook

| Notebook | Failing cell | Cause | Fix | Re-run |
|---|---|---|---|---|
| `05_evaluation` | 26 | `MAX_SYMBOLS: 5` sat below the notebook's `MIN_PERIODS = 10`, so no date could carry enough symbols, `eval_summary` came out empty, and `make_subplots(rows=0)` raised | Raised the cap to 20 in `tests/overrides.yaml`. Guard untouched | **PASSES**, 54s isolated; 20 symbols, 1,995 dates, correctness gate 64 PASS / 6 FAIL |
| `08_tabular_dl` | 12 | `RuntimeError: Training completed but registered checkpoints are incomplete for tabm_s` | not started | red |
| `11a_pca` | 8 | `ValueError: temporal artifact not aligned with canonical CV splits: fold 0 validation 231/247`. Writer `etfs/04:118` derives folds from `prices`; reader `utils/modeling.py:596` derives from `labels`. They differ by exactly the 21-session label horizon | diagnosed, not applied - moves production numbers | red |
| `11b_ipca` | - | `Timeout waiting for execute reply (300s)`, reproduced at load 3.8, so genuine rather than contention | not started | red |
| `18_strategy_analysis` | 36 | `RuntimeError: No holdout backtest matches val rank-1 (57f770b1195a) strategy spec` | not started | red |

## Withdrawn - measurement artifacts, not defects

| Notebook | Reported as | Actually |
|---|---|---|
| `03_financial_features` | `hurst_100` point-in-time leak (#203/#260) | Fails at load 60-76, passes at load 14. Five reproductions - synthetic, `.over("symbol")`, real 56-ETF panel, full `per_entity_features`, and the exact seal assertion - all give max diff **0.0** |
| `07_gbm` | LightGBM CUDA failure | Local only. The guard probes `torch.cuda.is_available()` and hands the answer to LightGBM. CI has no GPU, takes the fallback, passes |
| `13_model_analysis` | join key `datetime[us]` vs `datetime[ms]` | Passed once the worktree got its own venv. Shared-venv artifact |
| `09_dl_lstm`, `10_dl_tsmixer`, `12_causal_dml` | failures | All three `Timeout waiting for execute reply` under 3x CPU contention; pass at low load |

## Environment findings that cost measurement time

1. **Shared `.venv`** (until 20:39 UTC) resolved `utils/` from another lane's branch and carried different `ml4t-*` pins. Any pre-20:39 measurement is void.
2. **`/tmp/ml4t-pm-<notebook>.log` is not namespaced by case study** (`tests/pm_helpers.py:961`). Four case studies have an `05_evaluation`; concurrent lanes overwrite each other's diagnostics. This misled me once.
3. **CPU contention at load 60-76 on 24 cores** makes timeout failures unmeasurable, and - per `03` above - can make a *named exception* appear too. Exception type is not proof of a real defect.
4. **#290 is live**: `~/ml4t/test-data` has uncommitted modifications. HEAD's etfs fixture is 24 symbols; the working copy differs. Seed from HEAD via `git archive`.
5. **The squash merge of #478 breaks `git rebase origin/main`** for branches cut from `group/stage-01-05-assembly`: it replays all 33 sweep commits. Cherry-pick your own commits onto a fresh branch instead.

### DIFF
--- a/tests/overrides.yaml
+++ b/tests/overrides.yaml
@@ -1344,7 +1344,12 @@ case_studies/etfs/04_model_based_features:
   timeout: 300
 case_studies/etfs/05_evaluation:
   parameters:
-    MAX_SYMBOLS: 5
+    # MAX_SYMBOLS must stay above the notebook's MIN_PERIODS (10), the symbols a
+    # date needs before its IC enters the series. At 5 no date could ever qualify,
+    # so every date was dropped, eval_summary came out empty, and the first figure
+    # failed on make_subplots(rows=0). The fixture carries 56 symbols; 20 leaves
+    # margin for dates where some are null. Measured at 112s against the 300s cap.
+    MAX_SYMBOLS: 20
   timeout: 300
 case_studies/etfs/06_linear:
   parameters:


stefan-jansen:
## Superseded by `cs/stage-05-etfs`, and the withdrawn hurst finding was real

Two things, checked 2026-08-09 against the branches rather than against `main`.

### 1. The one change here is tuned to a notebook that no longer exists in that form

This PR changes only `tests/overrides.yaml`, raising `case_studies/etfs/05_evaluation`'s
`MAX_SYMBOLS` from 5 to 20, on the diagnosis that "`MAX_SYMBOLS: 5` sat below the notebook's
`MIN_PERIODS = 10`".

`cs/stage-05-etfs` rewrote that notebook. On the branch there is no `MIN_PERIODS` at all, and the
parameter now reads:

```python
MAX_SYMBOLS = 0
...
if MAX_SYMBOLS > 0:
    top = eval_panel.group_by("symbol").len().sort("len", descending=True).head(MAX_SYMBOLS)
```

So the constraint this PR works around is gone, and 20 is a value derived against code that was
replaced. What the override should be now has to be decided against the rewritten notebook, and the
likely answer is the one `case_studies/cme_futures/05_evaluation` already documents for the same
class of problem: no `MAX_SYMBOLS`, because a reduced panel scores nothing under a cross-sectional
coverage gate and the reduction buys little.

That decision belongs to the stage-05 PR, which owns the notebook. **Recommend closing this one
unmerged.** Three streams currently edit `tests/overrides.yaml` - this PR, `infra/ci-chapter-fixture`
(the `NQ100_SYMBOLS` removal), and the stage-04 PR, which has to add four case-study entries no
stage-04 branch carries. Landing an override here that the stage-05 PR then has to re-derive adds a
conflict for no gain.

### 2. "`03_financial_features` hurst_100 - measurement artifact, not a defect" is refuted

The Withdrawn table says this "fails at load 60-76, passes at load 14" with "five reproductions all
give max diff 0.0". Load is not the mechanism, and the gate was right.

The failure is deterministic on a cold numba cache, and it reproduces single-threaded in a quiet
process. Full measurements are on ml4t/agent-workspace#315; the short version, all cold-cache runs of
the same test in the same worktree on the same data:

| `hurst_exponent_nb` variant | result |
|---|---|
| `@jit(cache=True)`, as shipped | FAIL, `Cell 37 ... ['hurst_100']` |
| `@jit(cache=False)` | FAIL, identical |
| `cache=True` + one kernel call before `rolling_map` | PASS |

The reason the standalone reproductions gave 0.0 is that they ran second, against a cache the first
run had already warmed. CI containers always start cold, which is why the job is permanently red
rather than intermittently red.

This matters beyond bookkeeping: `assert_values_agree` is the whole of D.3, and #315 was filed
warning that a gate which appears to raise at random will eventually be read as noise and worked
around. Recording it as a load artifact is that outcome. The fix is a one-line change in
`ml4t-engineer` plus a release and a `uv.lock` bump.

stefan-jansen:
## `11a_pca`: the writer/reader frame difference is real, but it is not what fails

The table above attributes `fold 0 validation 231/247` to `04_model_based_features.py:118`
deriving folds from `prices` while `utils/modeling.py:596` derives them from `labels`. I measured
all three derivations for `etfs` today before acting on it:

```
writer  generate_cv_splits(prices,      21D)  train_start 2013-01-03  val 2022-12-29..2023-11-29
reader  generate_cv_splits(labels_21d,  21D)  train_start 2013-01-17  val 2022-12-29..2023-11-29
reader  generate_cv_splits(labels_5d,    5D)  train_start 2013-01-07  val 2022-12-29..2023-12-21
```

`prices` versus `labels` moves **only `train_start`**, by the label horizon truncating the head of
the label series. The validation window is identical. `validate_temporal_fold_coverage` allows a
leading train-window gap up to `MAX_TEMPORAL_WARMUP_FRACTION` and allows validation none, and the
failure is on a validation window, so changing the writer's frame does not move that number.

What moves `val_end`, by exactly the 16 dates in the error, is the **variant label**.
`generate_cv_splits` ends the last validation fold `outcome_horizon` sessions before
`holdout_start`, so `fwd_ret_5d` at 5D gets a longer fold-0 validation window than `fwd_ret_21d`
at 21D. `04_model_based_features` writes one `fold` column, tagged for the primary label, and
`run_case_study_variants` scores `fwd_ret_5d` against it. The primary label measures 231/231 with
nothing missing, which the frame reading cannot account for.

So `11a_pca` needs the variant refused until `04_model_based_features` emits a fold column per
outcome horizon, which is a stage-04 change requiring a production re-execution. Tracked
separately; the guard is on `infra/ci-fixture-registry` at `99789c61`.

## Two other rows here are now settled differently

**`03_financial_features` / `hurst_100`.** The withdrawal in this PR is based on five
reproductions giving max difference 0.0, and I reproduced the failure instead. It is not load and
not a leak: it is the **first build in a process with a cold numba cache**. Warm cache, 12 trials,
0 disagreed. Cold cache, 3 trials, 3 disagreed by 0.12 to 0.28. Capping BLAS and OpenMP threads
does not change it, and pinning polars to one thread does not change it; compiling
`hurst_exponent_nb` once before the build removes it completely. In one run 599 rows were wrong,
all in one symbol of 40, that symbol's whole column frozen at the first window's value. The cache
lives in `site-packages`, so it is cold exactly once per environment, which is why a repeat run
passes. `assert_values_agree` was right both times.

**Finding 4, `~/ml4t/test-data`.** Fixed. The working copy's `sp500_options` registry carried
5 training runs / 6 prediction sets / 68 backtests against HEAD's 12 / 13 / 88; it is restored to
`origin/main`, and the seed in `tests/conftest.py` now replaces what it seeds rather than keeping
whatever a previous run left, so re-running into an existing `ML4T_OUTPUT_DIR` no longer
accumulates registry rows. `git archive` from HEAD is no longer needed.

**Finding 2, `/tmp/ml4t-pm-<notebook>.log` collisions across case studies**, is a real defect and
is not fixed here.

stefan-jansen:
Re-measured on the current phase-0 base on 2026-08-12, as requested in the adjudication: `PYTHONUNBUFFERED=1 uv run pytest tests/test_case_studies.py -v -k 'etfs and 05_evaluation'` passes with the existing `MAX_SYMBOLS=5` override (1 passed, 176 deselected, 30.98s). The old empty-summary failure no longer reproduces because the notebook no longer has that MIN_PERIODS constraint, so raising the fixture universe to 20 would add runtime without fixing a current defect. Closing without merge.

PR #474, opened 2026-08-03
Prefer monthly Binance archives in async OHLCV and refresh FX CV folds

Rezzaa13:
## Summary
- Patch Binance async OHLCV downloads to prefer monthly ZIP archives (with per-month progress logs) instead of flooding daily requests for multi-year ranges
- Refresh `case_studies/fx_pairs/config/cv_config.json` to the canonical newest-first fold numbering and `val_size`/`holdout_*` schema used by downstream FX notebooks

## Test plan
- [x] Run `python data/crypto/market/download.py` (or equivalent crypto download) for a multi-month hourly range and confirm monthly progress logs / fewer daily ZIP requests
- [ ] Confirm `case_studies/fx_pairs/config/cv_config.json` loads in `05_evaluation` / `06_linear` without key errors (`holdout_start` / fold boundaries)
- [ ] Spot-check that fold 0 is the newest validation window (2023) and aligns with `generate_cv_splits` newest-first numbering

PR #486, opened 2026-08-07
Fix/matplotlib semibold font warning

Rezzaa13:


stefan-jansen:
Closing this one as a duplicate of #474 and #487, not as a rejection of the work.

The changed files here are exactly the union of the other two PRs:

- #474 - `data/crypto/market/download.py`, `utils/binance_public_monthly.py`, `case_studies/fx_pairs/config/cv_config.json`
- #487 - `case_studies/utils/sequence_dataset.py`, `case_studies/utils/tabular_dl.py`, `tests/test_sequence_dataset.py`

There is no matplotlib font change in the diff, so the title does not match the contents either. Reviewing the two narrower PRs separately is easier and lets each land on its own merits, so those are the ones we will work through. Nothing is lost by closing this.

Both #474 and #487 remain open.

PR #499, opened 2026-08-09
Two shared defects the stage-04 panes found and were not allowed to fix

stefan-jansen:
Eight stage-04 panes each finished with a report of what they were told to escalate rather
than edit. Two were defects in shared code, and both are fixed here. Neither moves a
published number: the HMM pin makes an artifact reproduce that already had the right values,
and the holdout change affects only notebooks that are all still `todo`.

## A seeded HMM fit was not a reproducible one

`case_studies/utils/temporal.py::fit_hmm_kmeans_init` promised a fit that does not move with
the seed, and delivered one that moves with the ambient thread count instead. Floating-point
addition is not associative, so a parallel reduction sums in whatever order the threads
finish; both the k-means partition and the E-step likelihoods inherit that, and EM carries it
forward. At a fixed seed, on identical data:

| threads | log-likelihood | transmat |
|---|---|---|
| 1 | -12263.024967575566 | `33959b0d5556f82e` |
| 2 | -12263.024967575568 | `7ab670b7e4f573d7` |
| 8 | -12263.024967575591 | `83f9dea9538e15fa` |
| 16 | -12263.024967575573 | `e816535a3ddb9fae` |
| ambient | -12263.024967576186 | `9c576c85c82c70c4` |

Inside `threadpool_limits(1)` all five return `-12263.024967575566` and one transition matrix.

The fifteenth digit is not the point. The digest is what says a notebook reproduces, and the
etfs, fx_pairs and crypto_perps_funding stage-04 artifacts each hashed differently run to run
because of this. Three panes reproduced it independently against their own notebooks and each
pinned it at its own call site; the docstring said pinning it was the caller's business, which
is what left six call sites to remember.

`threadpoolctl` was already installed as a scikit-learn dependency. It is imported here
directly, so it is declared directly.

**Not closed by this PR:** `etfs` and `cme_futures` still carry a local copy of
`fit_hmm_kmeans_init` inside the notebook rather than importing the shared one, so they take
the fix only through their own call-site pin. Both are on unmerged `cs/stage-04-*` branches;
the stage-04 PR assembly checks it.

## The holdout retrain was dropping three years of training data

`utils/modeling.py::append_holdout_fold_if_needed` builds the fold for the one fit the sealed
holdout ever sees, and took its `train_start` from `mds.splits[0]`. `generate_cv_splits` steps
backward from the holdout boundary, so fold 0 is the most *recent* fold and carries the
*latest* training start. Measured on etfs:

```
splits[0].train_start   2008-01-02
earliest train_start    2005-01-03
```

So the holdout retrain trained on the shortest window of the set rather than everything before
the seal, and only on the holdout path, where no validation number would show it.

`case_studies/etfs/04_model_based_features.py` states this exactly in its CV Fold Setup prose -
"Indexing the list hands it the shortest window of the set, silently" - and the docstring cited
that notebook, by line numbers that no longer resolve, while doing the thing the prose warns
against.

The six nasdaq100 stage-06-plus notebooks are the consumers. All are `todo` and no holdout run
on record used this path, so nothing published moves.

## Tests

Both fixes get a test that fails against the old behaviour, verified by reverting each:

- `tests/test_temporal.py` varies the *outer* thread pool and asserts the two fits are
  bit-identical, so removing the limiter makes them diverge.
- `tests/test_holdout_boundary.py` builds a newest-first fold set and asserts the appended fold
  starts at the earliest `train_start`.

Both files already run in the required `test-unit` job. 95 passed, 2 skipped across
`test_temporal`, `test_holdout_boundary`, `test_leakage_detectors`,
`test_modeling_input_lineage`, `test_notebook_sync` and `test_notebook_output_hygiene`; `ruff`
clean.


stefan-jansen:
Correcting a claim in the PR description, and answering the review finding.

The description says etfs and cme_futures "still carry a local copy of `fit_hmm_kmeans_init`
inside the notebook rather than importing the shared one". That is true of `main` and false of
the branches those notebooks are actually on. Measured on all seven pushed `cs/stage-04-*`
branches: **zero** local copies of any shared temporal helper, and etfs, cme_futures,
crypto_perps_funding and fx_pairs all import from `case_studies.utils.temporal`.

```
etfs         89: from case_studies.utils.temporal import (
cme_futures  85: from case_studies.utils.temporal import (
```

So the review's finding - that the pin "does not reach the artifact-producing workflows" - is
answered by the stage-04 branches rather than by this PR: each notebook pins at its own call
site *and* imports the shared helper, so once stage 04 merges there is one implementation and
one pin. What this PR removes is the requirement that every future call site remember, which is
what the docstring had made "the caller's business".

The three that do not import it (sp500_options, sp500_equity_option_analytics,
us_equities_panel) fit no HMM.

The duplication that genuinely remains is on the GARCH path, not the HMM one:
`fit_then_filter_garch`, `write_model_based` and `validation_only_slice` are described in four
separate pane reports as code each notebook carries its own copy of. Extracting them is tracked
against the stage-04 checklist rather than added here, because writing shared code from inside
one pane is what the split was avoiding.

PR #511, opened 2026-08-10
Take ml4t-data to 0.1.2 and un-break the two chapter 02 notebooks

stefan-jansen:
#510 moved every `ml4t-*` library to its stable release and knowingly left two chapter 02 teaching notebooks failing against the new pins, recorded in that commit's message. Both needed library changes rather than notebook changes, which `ml4t-data` 0.1.2 carries.

## What this does

**`13_data_quality_framework`** - the anomaly detectors rejected a `pl.Date` timestamp column, so every detector cell raised. 0.1.1 accepts it. Separately, the notebook asked `OHLCVValidator` for `check_negative_prices=True`, which the library migrates to `negative_price_policy="forbid"` and warns about, so the executed page showed the reader a `DeprecationWarning` naming a different API than the cell demonstrates. `rules.py` maps `True` to `"forbid"`, so no validation result moves.

**`18_data_management`** - two defects, both in the library.

`HiveStorage` encodes the partition key for filesystem safety and commits each write into a new generation directory, so the store's layout is no longer recoverable by globbing its root: the notebook's `rglob("*.parquet")` walk found the files but could not read the symbol or the period back out of the path. 0.1.1 adds `HiveStorage.partitions()`, which reports `label` and `size_bytes` per partition, and the two cells that inspected the directory tree now ask the store instead.

The cell captioned "Verify data is current" then printed `last updated: None` for all three symbols. The value was on disk; `normalize_storage_metadata` merged the provider's `custom` block over the storage record wholesale, and the incremental `update()` path writes a block whose `last_updated` is null, so the null won over the record's own stamp. Fixed in 0.1.2 (ml4t/data#40, merge `d558dc6`):

```
AAPL  None -> 2026-08-10 16:49:30.467769+00:00
MSFT  None -> 2026-08-10 16:49:31.170910+00:00
GOOGL None -> 2026-08-10 16:49:31.785906+00:00
```

## Verification

Both notebooks execute clean under the new pins - 0 errors, 0 unexecuted cells, 0 deprecation warnings - and are re-stamped. AAPL reports 24 partitions over the two-year load, one per calendar month, which is what the prose under that cell already said the layout would be.

0.1.2 also declares `Requires-Python <3.15,>=3.12`.


stefan-jansen:
Held as a draft until `ml4t-data` 0.1.2 ships. This PR leaves `18_data_management` printing `last updated: None` under a cell captioned "Verify data is current"; [ml4t/data#40](https://github.com/ml4t/data/pull/40) fixes that at the source. Both land together, with the pin moved to 0.1.2 and 18 re-executed.

PR #599, opened 2026-08-24
Crypto modelling stage: a placebo test that can fail, and the test that pins it

stefan-jansen:
The modelling stage of the crypto perpetuals funding case study: `07_gbm` through
`12_model_analysis`, plus the two chapter-15 DML notebooks that share the same
causal code.

## The placebo test could not fail, and neither could the test pinning it

Three defects in the block-permutation refutation, in `case_studies/utils/causal.py`:

**The block was sized by the label horizon alone.** On an 8-hour label over
8-hour bars the horizon, the embargo and the cadence all coincide at one bar, so
the "block" permutation was a full within-symbol shuffle - the exact degradation
`block_permute`'s own docstring names as making a placebo test too easy to pass.
The treatment is `premium_zscore_14d`, a 42-bar rolling statistic, and
`15_causal_estimation/04_dml_crypto_regime` already sweeps 3/21/42 for that same
treatment on that same data, headlining 21 because "a 3-bar (24h) block preserves
only short-range autocorrelation". The resolver was handing the placebo a third
of the block that notebook calls inadequate.

Two scales create the serial dependence the placebo has to preserve - the
overlapping labels span the horizon, the treatment spans its own construction
window - and the block now spans the longer. The window is read from
`features.windows` in `setup.yaml`, which already declares every window once, so
the number is not parsed out of a column name. For crypto this takes the block
from 1 to 42.

**The p-value omitted the plus-one correction**, so a 100-permutation run
published `p = 0.000`, a value no finite number of permutations supports.

**The pass/fail label was emitted at placebo counts that cannot produce it.**
With the correction, the smallest attainable p-value is `1/(n+1)`, so below 20
successful placebos the label read "Fails" whatever the data showed. It now
reports `Below resolution`.

### The test that named the regression and could not fail

`test_the_placebo_block_spans_the_label_horizon` asserted
`block_size == expected_block` across three label buffers. But the resolver
derived `embargo` and `horizon_steps` from the same expression, so reverting the
change it named passed all three parametrizations. Its own docstring conceded the
condition ("passes only while the embargo happens to be derived from the same
buffer") and that condition held unconditionally on the code path.

The replacement varies the horizon and the treatment window independently, and
was mutation-checked: restoring `block_size = horizon_steps` fails the
8h-horizon/21-bar-treatment case. This is the most transferable thing on the
branch - a fixture that agrees with itself proves nothing.

## What the refutation now says

`11` and `12` are re-executed. The causal identity moves from `401a87fb2169` to
`4a7d323f9c80`.

The refutation p-value does not move: `0.009901` before and after. That is worth
stating, because it was read during review as the signature of a placebo
distribution collapsed by the iid shuffle. It is not - with a 42-bar block, no
placebo reaches the observed effect either. `0.009901` is `1/101`, the floor of a
100-permutation test, and `12` now says so instead of reporting the floor as
though it were a measurement.

`12` also reports the estimate against its own standard error: `+0.001032` with a
HAC standard error of `0.0009709`, a ratio of `1.06`. The placebo test asks
whether the procedure manufactures the effect out of permuted treatment; the
standard error asks whether the effect is separable from zero. This estimate
survives the first and fails the second, and the notebook says what that
combination is and is not worth.

`12` also says what the ratio is not. The placebo block spans the treatment's
42-bar window, but the second-stage HAC bandwidth does not: it is the
cube-root-of-decision-times default raised only to cover the label horizon
(`utils/causal.py:467-472`), which is one bar here, and nothing ties it to the
treatment. Which way that moves the standard error is not something the mismatch
settles - a HAC estimate is not monotonic in its bandwidth, since the
autocovariances a longer one admits can carry either sign - so the notebook says
to read the ratio as provisional until the estimate is recomputed across a range
of defensible bandwidths. The realized bandwidth is not printed because it is not
registered - `causal_runs` stores neither `hac_maxlags` nor `n_periods` - and
adding a column to a shared registry is the same out-of-scope change as the
supersession one below, so the notebook states the rule instead of a number it
cannot read back.

## Which case studies this reaches, and which it does not

Nine case studies have a causal stage, but they do not share one entry point,
and the difference decides who the fix reaches.

Three resolve through the registry (`study.causal(...)` -> `resolve_causal_request`):
crypto, `cme_futures`, `fx_pairs`. Those are the only ones that get the new block
sizing. Crypto declares its treatment's window and gets 42 bars. The other two do
not - `carry_pct` matches none of `carry_smoothing`/`carry_zscore`/`carry_momentum`,
and `mom_skip_recent` matches neither `momentum` nor `skip_recent` - so they fall
back to the label horizon and emit a `UserWarning` naming the treatment.
Deliberately a warning rather than a silent default, because a silent fallback
would make them look fixed when they are not.

The remaining six - `etfs`, `us_firm_characteristics`, `sp500_options`,
`nasdaq100_microstructure`, `sp500_equity_option_analytics`, `us_equities_panel` -
call `run_dml_analysis` directly and never enter the resolver. Every one of them
passes `block_size = embargo_periods`, which is the original defect this PR fixes
for crypto, sized off the embargo rather than off anything the treatment does.
They are untouched here and they do not warn, because the warning lives in a
resolver they do not call. Six notebooks still size their placebo block the wrong
way, and this PR does not fix them.

Declaring windows would not be uniform work either. `features.windows` is
suffix-keyed on crypto (`premium_zscore: {14d: 42}`) but holds bare ints
(`skip_recent: 21`) and lists (`momentum: [5, 10, 21, ...]`) on etfs and
sp500_options. The lookup returns None and warns for those shapes rather than
guessing which list element built the treatment - the first version called `.get`
on whatever it found and would have raised `AttributeError` inside the shared
resolver for two case studies that had declared nothing wrong.

## It does something to two of them that has to be dealt with

The change is identity-bearing twice over. `ResolvedSpec.identity` hashes
`computation` in full, and the four new `refutation` keys - `block_size_basis`,
`label_horizon_steps`, `treatment_window_steps`,
`temporal_gap_tolerance_cadences` - live inside it. But the keys are not even the
operative cause: `computation["source_identity"]` is
`{"causal.py": sha256(case_studies/utils/causal.py)}`, so *any* edit to that file
moves every causal identity, and a comment would have done it. The causal hash
therefore moves for every case study routed through `resolve_causal_request`, not
only for the one whose block size actually changed, and there is no version of
this fix that avoids it.

Two have production notebooks publishing the old hashes:
`cme_futures/11_causal_dml.ipynb` shows `05b4cdb8a2f2` and `f78daa857772`, and
`fx_pairs/11_causal_dml.ipynb` shows `6e17a9b4644c`, `e9623aa44d9a` and
`f53540351b6b`. Those no longer reproduce. Only crypto's `11` and `12` were
re-executed here.

Re-running `11` alone does not clear it either: with no supersession on
`causal_runs` (below), a re-run writes a second canonical row for the same label
and the next notebook fails with `resolved to 2 identities`. Crypto hit exactly
this, and the stale row was deleted by hand after backing the registry up - a
workaround, not a fix, and the same manual step would be needed for both of the
others. Doing that across two case studies with their own branches open is
coordination this PR cannot do on its own.

## The notebooks describe the run

- `08`-`10` described a strategy-spec shape the resolver no longer accepts, and
  `08` carried two dead fallbacks in the notebook path.
- `07_gbm` named a fold-preparation path it does not take.
- `08`-`12` are re-produced against the run log, which a reset had left behind.
- `11` puts `block_size` and `block_size_basis` on its spec table rather than
  leaving them in the spec, since the block size is what the refutation rests on.

## The interface proof runs again

`scripts/prove_crypto_interface.py` imported `resolve_model_requests`, since
renamed to `model_requests` with a changed contract, so the proof died on
`ImportError`. It runs again and proves more: 12138 eligible rows, 6222 fills,
3366 trades, 2187 funding events.

`ETF_RELEASE_ROOT` also hardcoded a second checkout at `~/ml4t/public`, which made
the whole proof raise `FileNotFoundError` on any machine without that exact path.
It defaults to this checkout, takes `ML4T_ETF_RELEASE_ROOT`, and skips the
cross-study isolation check with a printed note when the root holds no ETF
predictions rather than failing outright.

Run it as a module - `uv run python -m scripts.prove_crypto_interface <workspace>` -
because `python scripts/...` puts `scripts/` on `sys.path` instead of the repo
root, and `research/workspace.py` imports `scripts.create_experiment`.

## Known gaps, deliberately not fixed here

- **The training loop prints an absolute save path**, which ships the author's
  worktree name inside committed notebook outputs. Fixing it changed the sha256
  of `deep_learning.py`, which `_sequence_source_identity` digests whole, which
  moved every sequence-model training hash and made `12` refuse to run against
  the frozen population. Reverted and confirmed by digest. It belongs with the
  next deliberate refit of 08/09/10, not as a side effect of a cosmetic change.
- **Causal results have no supersession.** `CausalResult.one` requires exactly one
  non-preview identity per label, so a refit that moves the identity registers a
  second and the next notebook fails with "resolved to 2 identities". Populations
  and candidate sets both have supersession; `causal_runs` has no equivalent. Any
  reader who re-runs a causal notebook after a code change hits this. Adding a
  column to a shared registry was out of scope with other branches in flight.

## Verification

`test_causal_adapter.py`, `test_causal_panel_splits.py`,
`test_causal_refutation_pvalue.py`, `test_research_workspace.py`: 64 pass, plus
the 18 in the refutation suite. Parameter-binding checks over the changed
notebooks: 48 pass. `11` and `12` executed in production with provenance stamped.

The backtesting stage (`13`-`17`) is a separate PR stacked on this branch.


christy108:
I've tried unsubscribing but I still get these emails. Please remove me from this list.

PR #604, opened 2026-08-24
Equity-option analytics stages 06-18, and a holdout leak in the shared causal resolver

stefan-jansen:
Stage-06-through-18 review pass for `sp500_equity_option_analytics`, plus the shared-library fixes that pass turned up. Nothing in this branch has been executed in production: `training_runs` holds only `gbm`, `linear` and `tabular_dl` rows from earlier work, and `causal_runs` is empty.

## The one that reaches other case studies

`resolve_causal_request` set the development cutoff as `holdout_start - buffer_delta`, subtracting **calendar** time from a boundary whose labels are counted in **trading sessions**. Every closure inside the buffer window - a weekend, a holiday - left sessions on the analysis side whose labels resolve after the holdout opens.

The cutoff now steps back a counted number of observations, taken from the pre-holdout axis and measured at the panel's own cadence. Two peer sessions measured the exposure on their own panels rather than taking my arithmetic for it, and both corrected it: `cme_futures` leaks 2 and 7 sessions on its 5d/21d labels, `fx_pairs` 1/3/8 sessions (20/60/160 rows), and it is *every* label, not only the long-horizon ones, because both holdouts open on 2024-01-01 and the New Year closure sits where an ordinary week was assumed. Their published causal estimates carry that contamination today; this branch's do not, since none exist yet.

`12_causal_dml` carried the same defect independently: its old cutoff left 7 sessions against a declared 10-session buffer. Now 2020-12-17.

## Buffer and outcome horizon are two quantities

The buffer keeps a fold's training rows clear of its validation labels. The outcome horizon is how long one outcome stays open. Only the second describes the overlap Newey-West corrects for, so a bandwidth derived from the buffer is right by configuration wherever it is right at all - and Newey-West is not monotonic in the bandwidth, so being wrong about it is not an error in a known direction.

`main` derives its horizon from `mds.label_buffer` and records that as `estimand.outcome_horizon`, which is why this branch keeps its own version of that one thing after resolving everything else toward `main`. Here the per-entity seal and the placebo block still take the buffer, which is the conservative side of the two; the bandwidth and the recorded horizon take the declared label horizon through `resolve_label_horizon`. A horizon longer than its buffer is refused, naming `labels.buffer`, `labels.variant_buffers` and `labels.horizons`. Two tests cover it on a fixture where the two actually differ: a 24H buffer over an 8H label on 8-hour bars is three bars against one.

They were conflated in the HAC bandwidth, in `run_dml_analysis`'s contract, and in the notebook. The bandwidth must cover the **label horizon** (the overlap it corrects for); the CV block must cover the **buffer**. Newey-West is not monotonic in bandwidth, so a wrong one is not merely a power cost in a known direction.

`resolve_causal_request` now reads the horizon through `resolve_label_horizon` and raises, naming `labels.buffer` / `labels.variant_buffers` / `labels.horizons`, when a declared horizon exceeds its buffer. The notebook applies the same rule at its own boundary.

## CI coverage that never ran

- `09_dl_lstm` / `10_dl_patchtst` were held out of the matrix by `gpu: true`. Measured: they run on CPU in ~40s under a 5-symbol preview. Both now have real CI entries. The reason recorded for the exclusion (an irreducible 60-observation lookback) was not what was keeping them out.
- Three pre-existing latent-factor defects: an `ipca` `checkpoint_interval` the adapter rejects outright, `11_latent_factors` reading the production registry instead of the preview root, and an aggregator collection order that put the bare aggregator ahead of its lettered siblings in six case studies.

## Verification

- `tests/test_causal_adapter.py`, `test_causal_panel_splits.py`, `test_causal_refutation_pvalue.py`, `test_causal_registry_registration.py`, `test_deep_learning_adapter.py`, plus the merged-in `test_coverage_gate.py` / `test_pm_helpers.py`: 173 passed at the merge commit.
- Previews: `09` and `10` exit 0, `09`'s second run printing `0 trained, 6 read` (the reuse fix, end to end); all five fitting latent-factor siblings exit 0.
- Sequence-family memory measured at 3.7 GB (20 symbols) and 4.6 GB (200), dominated by fixed overhead, extrapolating to ~6.8 GB on the full 630-symbol panel.

## What this PR does not do

No fitting notebook has been re-executed. Three shared files hash themselves into model identities - `causal.py`, `deep_learning.py` and `latent_factors/adapter.py` - and carry pending edits on other branches, so a production run before those land would be paid for twice. The freeze is per re-execution, not per merge: merging this is safe, and previews are carved out by design, since a preview tier cannot create an official population.

`measured` is deliberately unclaimed for `latent_factors`: those decompose the cross-section, so 20-symbol memory does not extrapolate the way the sequence family's does.

---

## Rebased onto main, and what that changed

`main` moved six commits under this branch while it was open. Two of them replace work that was here:

- **#606's per-entity seal replaces this branch's `pre_holdout`.** Mine stepped back over `.select(date_col).unique()` - distinct panel timestamps - while a label advances by the entity's own observations, so an entity absent from some of the final sessions kept rows resolving inside the holdout. `cme_futures` measured six days of difference on a real panel. Tracked as ml4t/agent-workspace#908; the construction is gone from this branch and `main`'s stands unmodified. What survives here is the pair of test updates that fix needs: `test_canonical_causal_uses_the_full_declared_population` and `test_canonical_causal_ignores_a_sample_cap_declared_in_the_preset` were asserting a property of the continuous eight-hour fixture rather than of the seal.
- **#603's "Underpowered" replaces this branch's placebo floor.** Both derive the same threshold of twenty draws from the plus-one correction. Refusing anything below it takes the preview tier's ten-draw runs out entirely; reporting says what those draws can and cannot show. `main`'s is the better shape and this branch's floor and refusals are dropped.

Kept alongside them: `max_symbols` in the causal identity, and `case_dir` forwarded so a run is written where its execution tier says rather than where the process happens to sit.

## One defect this turned up in a neighbour

`case_studies/etfs/12_causal_dml` asked for five placebo draws, below the ten `run_dml_analysis` assembles a refutation from at all. It published none, and the notebook reads `ref.get("empirical_p", 1.0)`, so `p = 1.00` was written to `causal_runs.refutation_p_value` alongside a `refutation_class` of "Fails" - a measurement from a test that never ran.

Every on-screen trace of it is suppressed. The printed refutation line is guarded by `if ref:`, and the placebo histogram and the subtitle carrying the p-value are inside a `len(placebo_arr) > 0` block, so someone watching the notebook run sees no line, no figure and no error while the registry gains a row nothing downstream can distinguish from a hundred-draw refutation that genuinely failed. A value that is never displayed cannot be sanity-checked by whoever ran it, which is why the count is raised here rather than the default improved: twenty-four draws, measured at 18s against a 120s cap, removes the trigger.

The `1.0` default is the actual defect, in three notebooks on `main`, and is filed as ml4t/agent-workspace#914 rather than fixed from this branch. This case study's own notebook raises instead of defaulting.


stefan-jansen:
## Known defect in this branch's cutoff fix - read before merging

The session-counted cutoff this PR introduces is correct about *sessions versus calendar days* and still wrong about *whose* sessions.

`case_studies/utils/causal.py` builds `pre_holdout` as `.select(mds.date_col).unique().sort(...)` and steps back `buffer_steps` over those **distinct panel timestamps**. A label advances by the entity's own observations. So on an unbalanced panel an entity that is absent from some of the final sessions has fewer than `buffer_steps` of *its* observations trimmed, and its last rows still resolve inside the holdout.

`cme_futures` measured the gap on a real panel: the panel-wide timestamp count seals at 2024-02-12, the within-entity count at 2024-02-06. Six days of rows the version in this PR admits.

**Nothing published here carries it.** `causal_runs` is empty for `sp500_equity_option_analytics`, so no estimate on this branch was produced under either construction. The defect matters for what merges, not for what this branch has already computed.

A balanced fixture cannot detect this - `cme_futures`' own tests did not, and their reviewer found it by reading the aggregation rather than by running anything. A test for it needs an unbalanced panel.

The fix `cme_futures` has extracted, which I am deliberately not duplicating here: compute the step-back within each entity, drop entities holding no more than `horizon_steps` pre-holdout observations since every row of theirs resolves past the boundary, then apply the earliest surviving cutoff panel-wide so the estimand keeps one scalar and the recorded number describes the filter actually applied.

Tracked as ml4t/agent-workspace#908, with the calendar-anchored-versus-session-counted horizon ambiguity one layer up as #909.

### Why this PR is still worth landing first

It carries the test updates that fix requires. `test_canonical_causal_uses_the_full_declared_population` and `test_canonical_causal_ignores_a_sample_cap_declared_in_the_preset` were asserting a property of the continuous eight-hour fixture rather than of the seal: on a continuous grid the trim removes nothing visible, so both passed for a reason neither test states. They are updated on this branch (`tests/test_causal_adapter.py:138` and `:368`); `main` still has the old versions at `:121` and `:346`.

Whoever lands a cutoff fix without this branch gets two population-test failures that look unconnected to a cutoff change.

PR #609, opened 2026-08-25
A causal refit had no way to retire what it replaced

stefan-jansen:
`CausalResult.one` resolves a label to exactly one current identity, and a refit under a changed `case_studies/utils/causal.py` produces a second. `causal_runs` had no supersedes column, no tiebreak and no recency rule, so the registry ended up holding two rows no reader could choose between - and the failure surfaced hours later, in a different notebook, with nothing saying which run caused it. `cme_futures` reached that state after re-running `11_causal_dml`: four canonical rows for two labels, and `12_model_analysis` exiting 1 with nothing to do but hand-edit the registry.

This mirrors `official_populations`, where a changed population under an existing name must name the hash it supersedes. Recency is deliberately not the fallback: `created_at` ties on a fast refit, it would be the only recency rule in a registry that is otherwise entirely spec-addressed, and it would hide the failure rather than surface it.

## What is here

- `causal_runs.supersedes_hash`, nullable, with a migration. Both reads probe for the column rather than naming it - neither goes through the migrating opener, so naming it would raise `OperationalError` on every registry written before today.
- `current_causal_identities` in `store.py`, used by the write-time check and by `CausalResult.one`, so the two cannot drift. The first draft had two copies and they disagreed within the hour.
- A new fit that would leave a label with more than one current identity is refused at the write, naming the one to retire. A repair step - a declaration on a row that already exists - may reduce the count without finishing, because chaining three stranded rows is incremental and requiring one at every step deadlocks.
- `declare_causal_supersedes`, the narrow fill-once write for the repair path. It is needed because re-running the newer identity reproduces its hash, so the fit is served from cache and never reaches the registering write.
- `supersedes` travels `study.causal(...)` -> `CausalRequest` -> `ResolvedCausalRequest` -> `run_resolved_causal_request` -> `register_causal_run`, alongside the spec and never inside it: the spec is hashed, so a declaration in it would change the identity of every run that carries one.
- `SUPERSEDES_CAUSAL` joins `SUPERSEDES_POPULATION` in `PRODUCTION_SAFE_PARAMETERS`.

## Cost

Threading the parameter through the notebook-facing wrapper edits `case_studies/utils/causal.py`, and `_causal_source_identity` hashes that file, so every `causal_hash` moves. Measured: **12 rows across four case studies** - `cme_futures` 4 (2 labels), `us_firm_characteristics` 3 (1 label), `fx_pairs` 3 (3 labels), `crypto_perps_funding` and `etfs` 1 each. The first two are already holding more identities per label than a reader can resolve, so they refit either way, and all five are inside the stage 06+ rebuild.

## Tests

`tests/test_causal_supersedes.py`, 19 tests. Nothing covered the two-identity path before - the four in `test_causal_result_read.py` are all about the refutation draw count, which is why this shipped. Both sides are covered: the write-time refusals, the reader's subtraction, an unchanged declaration re-registering cleanly, a retired row that cannot retire its own successor, fill-once surviving the migration backfill, the repair landing from cache, three stranded identities chained back to one, and a new fit into that same broken registry still refused. 54 passed across the causal and registry surface.

## Not here

No causal notebook declares `SUPERSEDES_CAUSAL` yet. Four need `SUPERSEDES_CAUSAL: str = ""` in the parameters cell and `supersedes=SUPERSEDES_CAUSAL or None` at the `study.causal(...)` call: `cme_futures/11_causal_dml`, `fx_pairs/11_causal_dml`, `crypto_perps_funding/11_causal_dml` and `us_equities_panel/14_causal_dml`. Each sits on its own `cs6` branch with a different version of that file, so the edit belongs to the case study that owns it, with its refit.

### DIFF
--- a/.github/scripts/notebook_provenance.py
+++ b/.github/scripts/notebook_provenance.py
@@ -145,6 +145,11 @@ PRODUCTION_SAFE_PARAMETERS: dict[str, object] = {
     # adds a declaration rather than removing work, so it cannot reduce the run, and
     # population.py raises unless it equals the current population hash exactly.
     "SUPERSEDES_POPULATION": VALIDATED_BY_CONSUMER,
+    # The causal identity this run retires. Same shape as SUPERSEDES_POPULATION and
+    # for the same reason: it adds a declaration rather than removing work, so it
+    # cannot reduce the run, and register_causal_run refuses a hash that is not a
+    # current canonical identity for the same label.
+    "SUPERSEDES_CAUSAL": VALIDATED_BY_CONSUMER,
 }
 
 
--- a/case_studies/research/causal.py
+++ b/case_studies/research/causal.py
@@ -42,19 +42,41 @@ class CausalResult:
         if not db_path.is_file():
             raise ValueError(f"causal selection for {label!r} resolved to 0 identities")
         with sqlite3.connect(db_path) as db:
+            # Same reason `open` below probes rather than naming the column: this read
+            # does not go through the migrating opener, and a registry written before
+            # supersedes_hash existed would raise OperationalError instead of
+            # resolving the single identity it does hold.
+            columns = {row[1] for row in db.execute("PRAGMA table_info(causal_runs)").fetchall()}
+            supersedes_column = (
+                "supersedes_hash" if "supersedes_hash" in columns else "NULL AS supersedes_hash"
+            )
             rows = db.execute(
-                "SELECT causal_hash, spec_json FROM causal_runs WHERE label = ? ORDER BY causal_hash",
+                f"SELECT causal_hash, spec_json, {supersedes_column} FROM causal_runs "
+                "WHERE label = ? ORDER BY causal_hash",
                 (label,),
             ).fetchall()
+        # A refit produces a second canonical identity for the same label. Which one is
+        # live is declared, not inferred: `created_at` ties on a fast refit, and a
+        # recency rule would be the only one in a registry that is otherwise entirely
+        # spec-addressed. A row another row names as superseded is retired.
+        retired = {row[2] for row in rows if row[2]}
         current = [
             causal_hash
-            for causal_hash, spec_json in rows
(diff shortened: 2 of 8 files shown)


stefan-jansen:
Two more things landed here since the PR was opened, both because they edit `case_studies/utils/causal.py` and this branch already pays for that file's digest change. Landing them separately would move all twelve causal identities a second time for nothing.

**The refutation threshold, from #914's shared-runner half.** `run_dml_analysis` computes the permutation test only at ten successful placebo draws and leaves `refutation` empty below that, silently. Walking every declaration in `tests/overrides.yaml`:

| declared as | sites | value | effect |
|---|---|---|---|
| `N_PLACEBO` | 2 | 5 | the test could never run |
| `N_PLACEBO` / `PREVIEW_N_PLACEBO` | 4 | 10 | one failed draw loses it |
| `n_placebo` (in `PREVIEW_REDUCTIONS`) | 2 | 10 | same |

Five of nine were below the threshold outright, so `refutation_p` registered NULL on every preview and CI execution - and the four notebooks that read it as `.get("empirical_p", 1.0)` published 1.0 as a measured p-value each time. All nine are now 15, the magic 10 is `MIN_PLACEBO_DRAWS`, and a test walks the file so it cannot drift back. Cost: 15 nuisance refits where there were 5 or 10, in the `cs-*` notebook jobs.

Enforcing the margin at request time was the first attempt and it was wrong - nine tests in `test_causal_adapter.py` request two to six draws on purpose, to exercise block-span and permutation feasibility without paying for a refutation they never read. The boundary is a property of a declared reduction, not of every call.

**`main` merged in**, so this picks up `test-unit-data` and `test-unit-image`, which are now required contexts.

Still not here, and per-notebook: the four notebooks that default the p-value instead of raising, which is `us_firm_characteristics`' behaviour and the better failure. And the `SUPERSEDES_CAUSAL` parameter cell.

PR #625, opened 2026-08-25
Build the holdout evaluation into the case study, out of the Chapter 20 notebook

stefan-jansen:
The holdout evaluation lived in the Chapter 20 notebook, so a case study could not
run its own final evaluation without the teaching notebook. This moves the
derivation into `case_studies/research/holdout.py` where the case studies can reach
it, with the contract the lock already requires.

**`build_holdout_cv()`** derives the holdout fold from the panel's own observations.
The buffer is counted in **observations stepped along the panel's dates**, never
subtracted as calendar time - 21D is about 15 trading days, and subtracting 21 days
walks the train end into the buffer and leaks. Month buffers cannot be divided by an
observation step at all, so they fall back to `periods_per_year`.

Verified against two independent references rather than against itself: the etfs
train end lands on 2023-11-30, matching what the causal resolver derives for the
same panel and buffer, and `us_firm_characteristics` comes out at one observation,
matching its notebook.

**`evaluate_holdout()`** refuses at the lock with a `NotImplementedError` naming the
fold-derived fields it cannot supply. `validate_locked_expected_keys` requires an
eligibility manifest describing the holdout frame, and building that needs a
per-family hook that does not exist yet. An explicit refusal is the honest state;
deleting the fields it cannot fill would break the reconstructors that require them.

`_normalize_label_buffer` is renamed `normalize_label_buffer` - it has callers
outside its module now.

16 tests: `tests/test_holdout_cv_derivation.py` (11) and
`tests/test_holdout_evaluation_driver.py` (5).

A first mutation check on this passed while the bug was live, which is worth
recording: replacing the step *count* left the tests green, because stepping back 21
observations is right whatever produced the 21. Only replacing the step-back
*mechanism* failed them. The tests were then written against the mechanism.

### DIFF
--- a/.github/ci/unit-test-quarantine.txt
+++ b/.github/ci/unit-test-quarantine.txt
@@ -51,6 +51,10 @@ tests/test_gbm_runtime.py
 tests/test_insight_chapter.py
 tests/test_latent_factors_no_leak.py
 tests/test_latent_input_identity.py
+# Imports `_install_fixture_adapter` and `_locked_study` from
+# test_locked_holdout_execution, so it inherits that module's torch dependency
+# transitively rather than importing torch itself. Same boundary, same reason.
+tests/test_holdout_evaluation_driver.py
 tests/test_locked_holdout_execution.py
 tests/test_model_analysis_selection.py
 tests/test_registry_metrics.py
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -947,6 +947,7 @@ jobs:
             tests/test_insight_chapter.py \
             tests/test_latent_factors_no_leak.py \
             tests/test_latent_input_identity.py \
+            tests/test_holdout_evaluation_driver.py \
             tests/test_locked_holdout_execution.py \
             tests/test_model_analysis_selection.py \
             tests/test_registry_metrics.py \
--- a/case_studies/research/holdout.py
+++ b/case_studies/research/holdout.py
@@ -469,3 +469,95 @@ def build_holdout_cv(
             "holdout_window": [str(holdout_start), str(holdout_end)],
         },
     }
+
+
+@dataclass(frozen=True)
+class HoldoutOutcome:
+    """One case study's holdout evaluation, and whether this call is what produced it."""
+
+    lock: ResearchLock
+    evaluated_now: bool
+
+    @property
+    def lineage(self) -> dict[str, str]:
+        return self.lock.study.lifecycle.holdout_lineage(self.lock.hash)
+
+
+def evaluate_holdout(
+    study: Any,
+    *,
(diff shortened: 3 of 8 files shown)


stefan-jansen:
**The eligibility hook is designed. Recording it here rather than implementing it tonight, because it touches five shared family modules while eight panes are running against them.**

## What blocks the lock

`evaluate_holdout` refuses at `_refuse_incomplete_holdout_spec` because `validate_locked_expected_keys` (`research/models.py:207`) requires `computation.expected_prediction_keys` describing the **holdout** frame:

```python
{"digest": value_digest(expected, ("symbol", "timestamp", "fold")),
 "n_rows": expected.height,
 "n_folds": expected.get_column("fold").n_unique()}
```

Building that needs the holdout window's eligible key frame, which only a family resolver can produce — it depends on the modelling dataset, the label's null/NaN validity rule, and the entity and date column names.

## The hook already exists in uniform shape

Every family exposes `reconstruct_locked_request(study, spec, *, checkpoint_kind, checkpoint_value)` — identical signature in `linear.py:618`, `gbm.py`, `tabular_dl.py`, `deep_learning.py`, and the latent-factor adapter. Each already derives `expected` from the single holdout split and calls `validate_locked_expected_keys` (`linear.py:675-682`, `gbm.py:1994-1995`, `tabular_dl.py:494`, `deep_learning.py:691`, `latent_factors/adapter.py:594`).

So the frame this needs is already computed by every family — just at execution time, from a spec that must *already* carry the manifest.

## The change

Extract the key-derivation half of `reconstruct_locked_request` into a uniform per-family entry point:

```python
def holdout_expected_keys(study: Study, spec: dict[str, Any]) -> pl.DataFrame
```

`evaluate_holdout` resolves it through the existing registry — `get_adapter("model", family)` in `research/adapters.py`, which already binds all five families — and populates `computation.expected_prediction_keys` before taking the lock. `_refuse_incomplete_holdout_spec` then narrows to the three genuinely fold-derived fields it names, which are re-keyed the same way.

**`reconstruct_locked_request` must call the same function**, not a parallel copy. One implementation per family, used at lock time and at execution time, so the manifest matches **by construction** rather than by two code paths agreeing — which is the failure this whole class of bug is made of.

## Why not tonight

Five modules under `case_studies/utils/` with eight panes executing against them. The change is additive, but a shared-module edit during live runs is the hazard two other panes correctly declined this evening, and a third measured: a mid-run source change registers rows whose provenance is false and **nothing catches it** (`memory/NOTEBOOK_EXECUTION_HAZARDS.md`, the branch-switch row).

The refusal in `evaluate_holdout` is the honest interim state — it names the exact fields and says why, rather than taking a lock that would fail at execution.

PR #636, opened 2026-08-27
Make notebook completion stable across non-semantic drift

stefan-jansen:
Semantic model identity versions, a verified identity migration, migrated fitted-state and schema validation, and final-checkpoint support.

A documentation-only or output-presentation edit no longer invalidates a computational result when semantic identity and registered inputs are unchanged. Only identities whose computation changed are re-run; presentation and provenance update without discarding valid model fits.

Five commits: e3784bda, 0d7e6263, b86fd78c, f7108e55, c1e1ed14.
Tests at push time: 98 model-adapter, 135 research/registry, 38 focused, 6 migration.

This is shared infrastructure under `case_studies/utils/` that eight case-study panes execute against, so it should land before the phase PRs that build on it.

### DIFF
--- a/.github/ci/unit-test-quarantine.txt
+++ b/.github/ci/unit-test-quarantine.txt
@@ -17,7 +17,7 @@
 
 # ---------------------------------------------------------------------------
 # Needs the modelling environment - run by test-unit-image, in ml4t/ml4t:latest.
-# 30 files. The 26 that existed on 2026-08-24 measured 392 tests, 3m21s.
+# 32 files. The 26 that existed on 2026-08-24 measured 392 tests, 3m21s.
 #
 # `case_studies/utils/gbm.py` imports torch at module scope, so the LightGBM
 # files are behind this boundary too even though LightGBM itself is a small
@@ -58,6 +58,7 @@ tests/test_latent_input_identity.py
 tests/test_holdout_evaluation_driver.py
 tests/test_locked_holdout_execution.py
 tests/test_model_analysis_selection.py
+tests/test_model_source_identity_versions.py
 tests/test_registry_metrics.py
 tests/test_research_contract_execution.py
 tests/test_research_model_planning.py
@@ -72,6 +73,7 @@ tests/test_sequence_identity_roundtrip.py
 tests/test_tabular_dl.py
 tests/test_tabular_dl_adapter.py
 tests/test_tabular_dl_runtime.py
+tests/test_training_identity_migration.py
 
 # ---------------------------------------------------------------------------
 # Executes notebooks through Papermill, against the reader's image and the
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -955,6 +955,7 @@ jobs:
             tests/test_holdout_evaluation_driver.py \
             tests/test_locked_holdout_execution.py \
             tests/test_model_analysis_selection.py \
+            tests/test_model_source_identity_versions.py \
             tests/test_registry_metrics.py \
             tests/test_research_contract_execution.py \
             tests/test_research_model_planning.py \
@@ -967,6 +968,7 @@ jobs:
             tests/test_tabular_dl.py \
             tests/test_tabular_dl_adapter.py \
             tests/test_tabular_dl_runtime.py \
+            tests/test_training_identity_migration.py \
             -q --tb=short --timeout=900 --timeout-method=thread -rf
 
   # ---------------------------------------------------------------------------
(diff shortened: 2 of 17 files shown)


stefan-jansen:
## Merge-order finding: this moves every published latent-factor and sequence identity once

Raised by the `sp500_equity_option_analytics` pane before merging its own PRs, and confirmed by
reading this branch.

**What changes.** `_source_identity` in `case_studies/utils/latent_factors/adapter.py` goes from a
dict of fourteen file SHA256s to `{latent_adapter: 1, latent_model: "<model>/v1"}`, and
`_sequence_source_identity` in `case_studies/utils/deep_learning.py` does the same via
`SEQUENCE_RUNNER_VERSION` / `SEQUENCE_PREPARATION_VERSION` / `SEQUENCE_STATE_VERSION` /
`SEQUENCE_BACKEND_VERSIONS` / `SEQUENCE_ARCHITECTURE_VERSIONS`.

Both dicts change shape entirely, so **every published latent-factor and sequence identity moves on
landing.** That is the intent - it is the permanent cure, and afterwards a cosmetic edit to either
file cannot move an identity, which is the defect this fixes.

**What it does not do.** `scripts/migrate_training_identity.py` takes `--case-study` and
`--source-training-hash`, both required, and migrates **one row at a time**. There is no sweep and no
per-registry driver. So the migration covers only what someone points it at.

**Consequence.** Landing this without running the migration leaves nine registries' worth of
published members no longer reproducing. Two case studies have quantified their exposure:

- `sp500_equity_option_analytics`: **947** prediction identities behind PRs #634 and #635
- `crypto_perps_funding`: **196** identities already moved once by cosmetic edits to
  `deep_learning.py` under #620/#604/#597 - deep_learning 100, tabular_dl 96, none present in
  `prediction_sets`

**This should still land** - it is the fix for the class, and crypto's 196 is exactly the failure it
prevents recurring. But a per-registry migration procedure has to exist first, and none does. The two
panes above have the counts and the need; the procedure is worth working out once and recording here
rather than derived nine times.

Related: ml4t/agent-workspace#631 established the same pattern for latent factors ("scope identities
to material inputs"); this generalizes it and makes it permanent.

stefan-jansen:
Concrete numbers from `crypto_perps_funding`, since this PR decides what my registry needs and mine is the case study where the cost of the current behaviour has actually been measured.

**What the whole-file digest cost, once.** `13_backtest` cannot run on `main` today. `freeze_official_model_population` re-derives the population from current source and refuses:

    ValueError: a changed population named 'crypto-validation-predictions-v1'
                must explicitly supersedes 21acd4cb3a8c

Measured against the registry:

| | |
|---|---|
| planned members | 677 |
| registered members | 677 |
| differing, each direction | **196** |
| families affected | deep_learning 100, tabular_dl 96 |
| families untouched | gbm 400, linear 81 |
| of the 196 new identities present in `prediction_sets` | **0** |

The split is exactly the families whose `source_identity` digests whole source files. Three commits on `main` since my merge-base touched that source: #620, #604, #597. #620 is "declare the device instead of inventing one" - the identities moved for an edit that changes no fit.

Without a migration the only way back is refitting 196 predictions and then re-running every backtest that references them, which is most of the funnel and every published number in stages 13 through 17. That is why this PR matters to me and not only in principle.

**What I would need from the migration.** `scripts/migrate_training_identity.py` takes `--case-study` and `--source-training-hash` and migrates one row at a time. My 196 are prediction sets rather than training runs, spread across the checkpoints of a smaller number of training runs, so what I need to know is:

1. Does migrating a training run carry its prediction sets with it, or does each prediction identity need its own call? The 196 are what `13_backtest` compares against, so the prediction hashes are what has to end up right.
2. Do the downstream rows follow? `backtest_runs.prediction_hash` is a foreign key and 2,031 signal-stage backtests point at these predictions. A migration that moves the prediction and leaves the backtest pointing at the old hash trades one broken invariant for another.
3. Is there a dry run? Before touching a registry that holds every published number for the case study, I want to see the list of rows it would rewrite.

**Offer.** If a sweep wrapper over the per-row tool is wanted, crypto is a reasonable place to work it out: 196 rows, four families of which two move, and an independent check available afterwards - `13_backtest` either runs or it does not, and it is the notebook that refuses today. `sp500_equity_option_analytics` has the same need at 947 identities, so it is worth doing once properly rather than twice by hand.

**One note on the identity move this lands.** The `source_identity` dict changes shape from file hashes to declared version numbers, so every DL and latent-factor identity moves on merge. Worth saying explicitly in the merge commit that this is the last such move and why - the value of the change is precisely that the next cosmetic edit cannot repeat it, and a reader looking at two consecutive identity moves has no way to tell that from the diff alone.

PR #639, opened 2026-08-28
Narrow the registry reads that decide a superseded hash to the failure they can answer

stefan-jansen:
Both guards that decide whether a notebook may offer a declared predecessor hash answered
every `sqlite3.OperationalError` with "nothing is published here". Only one operational
error means that: the missing table a clean clone has. A lock timeout, an I/O error and a
half-migrated schema reach the same `except` and get the same answer.

The consequence is not a wrong return value at the end of a cheap read. The predecessor is
withheld, the notebook refits every configuration - or, on the causal side, runs the DML fit
and every placebo refit - and then dies at registration for naming no predecessor. The
declaration exists precisely to avoid paying that twice.

The reads also took SQLite's five-second default, while every other reader of the same file
uses the 120s driver timeout and 60s `busy_timeout` that `_open_registry` sets. A registry is
written by notebooks, backfills and scripts at the same time, which is what makes a lock
timeout here an ordinary event rather than a hypothetical one - and it entered through the
`except` above without a trace.

`_connect` now sets those timeouts in one place for `OfficialPopulation.one`, `.open` and
`_lineage`. It cannot be `_open_registry` itself, which runs the schema DDL and would create
the very tables a clean clone is being asked about.

Filed by RoboRev as jobs #16098 and #16099 against `causal.py`; the same defect was in
`population.py`, which the reviews did not reach.

## Verification

Four new tests, each falsified against the reverted code with the revert confirmed present:

- reverting `population.py` fails `test_a_lock_timeout_propagates` and
  `test_the_read_waits_as_long_as_every_other_reader_of_this_file` (both connections opened
  with `timeout=None` pre-fix, asserted on the connection `population_supersedes` actually
  opens rather than on the helper in isolation)
- reverting `causal.py` fails its two counterparts (`busy_timeout` read back as 5000)
- the missing-table tests pass in both directions, which is the point: the clean clone is
  still answered with `None`

`tests/test_population_supersedes.py`, `tests/test_causal_supersedes_resolution.py`,
`tests/test_superseded_members_registries.py`, `tests/test_research_workspace.py`: 61 passed.

### DIFF
--- a/.github/ci/unit-test-quarantine.txt
+++ b/.github/ci/unit-test-quarantine.txt
@@ -86,14 +86,17 @@ tests/test_tabular_dl_runtime.py
 #
 # The first three collect nothing but notebook executions, so the whole file
 # goes. `test_model_registry.py` does not: only `test_model_notebook` runs a
-# notebook, and its other nine tests are pure config assertions that finish in
+# notebook, and its other ten tests are pure config assertions that finish in
 # 0.07s. Excluding the file threw those away, which is the same
 # exclude-more-than-you-meant defect the subtraction below exists to fix.
+# Ten, not nine: `test_registering_stage_maps_to_its_actual_entry_point` is
+# parametrized over five stages, so counting `def test_` undercounts what the
+# deselect leaves behind. Measured 2026-08-27: 88 collected, 78 deselected.
 # ---------------------------------------------------------------------------
 tests/test_case_studies.py                       # cs-<case-study> (test-case-studies)
 tests/test_chapter_notebooks.py                  # chNN (test-chapters)
 tests/test_docker_notebooks.py                   # test-py312, test-neo4j, test-benchmark
-tests/test_model_registry.py::test_model_notebook  # nothing runs it; the file's other nine tests run here
+tests/test_model_registry.py::test_model_notebook  # nothing runs it; the file's other ten tests run here
 
 # ---------------------------------------------------------------------------
 # Needs the reader's Docker image. It walks every declared third-party import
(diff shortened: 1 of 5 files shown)


stefan-jansen:
RoboRev's Medium finding on `population.py:326` is real, and it is the residual of the same defect rather than a separate one. Recording what it does and does not reach, so the next reader does not have to re-derive it.

**The scenario.** A preview study `P` with output root `W` activates, stamping `W/.preview`. A canonical study `C` sharing `W` then activates, stamping `W`. Asking `_preview_is_active(P)` now answers False - the stamp no longer ends in `.preview` - so `P` is offered the declared hash and `run_model_population` refuses the run for a preview attempting to supersede a snapshot.

**Why this commit still helps.** Before it, *any* active preview anywhere in the process withheld from *every* study, including canonical ones that merely ran second. That is the failure `#640` had to work around and the one that costs a backfill every fit before `create` refuses it. After it, the wrong answer needs two studies deliberately sharing one output root and activating in a particular order.

**Why the remaining hole is not closed here.** Both directions have the same root cause: `Study` does not retain its execution tier, so the tier is read from mutable process state whichever way you scope the read. The complete fix is to record the tier on the study at construction and read it from there, which retires `ML4T_OUTPUT_DIR` as a tier signal entirely. That touches `Study`'s shape, `activate`, `open_study` and the isolated-preview branch - which today carries no tier marker at all and relies on its registry being empty - so it wants its own change and its own review, not a fold-in here while several case studies are mid-run.

Tracked as ml4t/agent-workspace#950, which stays open for exactly this. `_refuse_preview_activation` keeps the unscoped process-wide question deliberately: "which tier is this study" is study-scoped, but "may a population be written right now" is not, because the write lands in whichever registry the active tier points at.

PR #650, opened 2026-08-28
Make the README callout date-free so it cannot go stale

stefan-jansen:
The blockquote above the book cover advertised a Navigate ML for Trading session on **July 30, 2026**. That event ran four weeks ago, so the most-read paragraph in a repo with 20,709 stars has been pointing every visitor at a past date.

Updating the date would fix today and reintroduce the same defect at the next event, which is exactly how this survived four weeks unnoticed. The callout now carries no date and no per-event Maven URL, and points at the courses page on the companion website, which is maintained as part of normal operations and already lists every live lesson, workshop and cohort.

The link keeps its UTM fields: Maven has no referrer field, so a GitHub arrival is currently indistinguishable from any other. Routing through the website, which does have analytics, is what makes this repo measurable as an acquisition channel.

One hunk, README only. Verified: no year or month name inside the callout, no `maven.com/p/<id>` URL anywhere in the README, and `https://ml4trading.io/courses/` returns 200.

Not merged - this is a marketing change to the public README, so it is yours to approve rather than a review-program PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,11 @@ strategy you can actually run, and keep running, in a live market.
 machine to a running notebook, prerequisites included. The short version is under
 [Quick Start](#quick-start) below.
 
-> **Free reader's guide:** Join [Navigate ML for Trading, 3rd Edition](https://maven.com/p/c6e0e7/navigate-ml-for-trading-3rd-edition)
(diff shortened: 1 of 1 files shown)


stefan-jansen:
Superseded by #651. That PR replaces this callout with a block generated daily from the Maven instructor profile, so it carries real dates and cannot go stale, rather than removing the dates to avoid the problem.

PR #672, opened 2026-08-30
etfs: remove the 5-symbol fixture cap that made every top_k scheme identical

stefan-jansen:
`tests/overrides.yaml` capped `case_studies/etfs/04_model_based_features` and `05_evaluation` at `MAX_SYMBOLS: 5` while the fixture price panel holds 24 symbols. Every scheme in the `top_k_grid` [5, 10, 20] therefore passed the feasibility filter and each one held the entire five-symbol prediction cross-section, so the notebook registered 12 backtests, 12 Sharpes and 0 trades - a sweep whose arms cannot differ.

Removing the cap restores the notebooks' own no-cap default (`MAX_SYMBOLS = 0`). Timeouts raised 300 -> 900 to cover the wider cross-section.

Unblocks etfs 14_backtest, 15_portfolio_management, 17_risk_management and 18_strategy_analysis (ml4t/agent-workspace#989).

Not fixed here, deliberately: `get_entry_schemes_for` (`case_studies/utils/sweep_config.py:367`) filters `k >= n_assets` against the price panel rather than the prediction cross-section. With a narrow cross-section the honest filter drops every scheme and registers nothing, which is a different failure. It wants fixing once the cross-section is wide.

Verified: 90 passed in `tests/test_pm_helpers.py`.

### DIFF
--- a/tests/overrides.yaml
+++ b/tests/overrides.yaml
@@ -1451,15 +1451,22 @@ case_studies/etfs/02_labels:
 case_studies/etfs/03_financial_features:
   timeout: 300
 case_studies/etfs/04_model_based_features:
+  # No MAX_SYMBOLS, for the reason 02_labels gives: the CI fixture is already only 24
+  # ETFs, so it IS the reduction. Cutting to 5 on top of it made every downstream
+  # backtest degenerate rather than fast. etfs declares top_k_grid [5, 10, 20] against a
+  # 24-fund price panel, so all three schemes pass the feasibility filter - but the
+  # predictions they rank covered 5 symbols, so top-5, top-10 and top-20 each held the
+  # whole cross-section, the basket never changed, and 12 of 12 backtests recorded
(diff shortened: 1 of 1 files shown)


stefan-jansen:
Superseded by #619. The entire contribution of this PR is a 17-line change to `tests/overrides.yaml` (12 insertions, 5 deletions), cherry-picked onto `cs6/etfs-band-c` as e38b1be5 with byte-identical patch content, verified against ea20129f before closing.

To be exact about why this did not turn `cs-etfs` green on its own: it is orthogonal to all three etfs CI failures. It fixes the degenerate top_k sweep (#989) - not the checkpoint identity mismatch in 09_dl_lstm and 10_dl_tsmixer, not the tsmixer dispersion guard (#988), and not the section 6 anchor in 18_strategy_analysis. That is not an argument against the change; four of the etfs skips cite it. It was simply never the thing that would flip the job.

Closing here so etfs carries one PR rather than two. The branch `fix/etfs-fixture-cross-section` can go once #619 merges.

PR #682, opened 2026-08-30
us_firm_characteristics/16: the executed holdout backtest, reproducing the same hash

stefan-jansen:
One notebook execution. #681 moved `16_holdout_backtest`'s conformal widths write
below the replacement guard, which changed a code cell, so its outputs and
provenance stamp were cleared and it merged claiming nothing.

This is that run. It reproduces `b2acfd5420c8` exactly - the same backtest hash
against the same holdout prediction set `6d996324b41a`, and one holdout row in the
registry rather than two. Moving the write changes when the artifact lands and
nothing about the specification, and the identical hash is what confirms that
rather than an argument that it should be so.

exit 0, largest single process 2.3 GB.

The registry now reads signal 1252, allocation 120, cost_sensitivity 22, holdout 1.


stefan-jansen:
Superseded. #681's squash merge diverged `cs6/us_firm_characteristics`, so this went DIRTY. The same single commit is now on `cs6/us_firm-16-executed`, cut fresh from main, carrying only the executed notebook.

PR #684, opened 2026-08-30
A holdout backtest states its contract instead of requiring a research lock

stefan-jansen:
`Strategy.__post_init__` required a `research_locks` row in the `LOCKED` state before it would execute any holdout backtest. That made an authorization token a prerequisite for a computation, and the cost was not the code: a holdout found to be wrong after a bug fix could not be corrected, because the lock is by construction the one artifact that cannot be revised. It also refuses correct holdouts produced outside the lock transaction - `fx_pairs`' rebuilt lineage is refused today by a lock recording a selection that no longer exists.

## What replaces it

`HoldoutExpectation` carries the facts the lock was standing in for, as an argument rather than a token:

- the retrain's **own** training hash - never the validation model's, which a genuine retrain does not share
- the checkpoint pair, which is the only thing separating two prediction sets of one training run
- `strategy_projection` of the selected validation strategy
- the validation prediction a conformal allocator calibrates its holdout widths from

**Both guarantees move, not just the first.** Matching the model alone would let a caller replay a different strategy on the correct model - a different route to the same wrong answer - so `_build_spec` compares the resolved holdout strategy against the expectation's projection.

**There is no default.** A holdout prediction with no expectation is refused. A fallback to "accept any canonical holdout prediction" would leave the constructor executing a holdout produced from a configuration nobody selected, which is the failure this exists to prevent.

## Evidence that the refusal is load-bearing

Replacing the refusal with a permissive branch fails exactly one named test - `test_a_holdout_without_an_expectation_is_refused` - and leaves the other five green. That mutation check is what distinguishes this from merely deleting the lock check.

The same check was applied to the derivation: mutating `HoldoutSelection.expectation()` to read the holdout prediction it gates, instead of the selection, fails the two derivation tests and leaves the other six green. An expectation built from the prediction would compare every value with itself and accept anything handed to it.

## The lock-free caller

`resolve_holdout_selection` derives the whole lineage from the immutable candidate set: the rank-1 validation backtest read from the set rather than accepted from a caller, the holdout training spec re-keyed onto the derived fold, and the training identity that retrain *will* have - computed before the retrain exists. That is the property a lock does not have: re-deriving after an upstream rebuild yields the current contract, not the retired one.

`StrategyReplay` no longer holds a `ResearchLock`; it holds the four facts the replay consumes, all readable from the selected validation backtest. `prepare_locked_strategy_replay` is a wrapper that states the lock's own contract, and `run_locked_holdout` is unchanged, so the locked path keeps exactly the guarantee it had.

`_locked_strategy_projection` moves from `lifecycle.py` to `strategy.py` as public `strategy_projection`. It was private to the lock module, which is why a holdout produced outside that transaction had no way to state the guarantee at all.

## Scope - which execution paths this actually covers

This binds the `case_studies.research` API path. The older `case_studies.utils.backtest_runner.run_backtest` path - which `us_firm_characteristics/16_holdout_backtest` uses - never had the lock check and does not gain this contract, so the two paths make different promises about the same holdout. Closing that is separate work; do not read "a missing expectation is a refusal" as covering both.

## Tests

105 pass across `test_holdout_expectation`, `test_locked_holdout_execution`, `test_research_contract_lifecycle`, `test_research_flow` and `test_holdout_cv_derivation`.

`test_case_study_pipeline[etfs-18_strategy_analysis]` fails on clean `main` for an unrelated reason (the `val_rank1_self` anchor cannot be resolved), verified in a worktree at `origin/main`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -93,6 +93,15 @@ jobs:
             # something that had already been tested in ninety seconds.
             shared:
               - 'utils/**'
+              # Every case study imports these two. `case_studies/research/` is the study,
+              # population, candidate-set and holdout machinery; `case_studies/utils/` is the
+              # backtest runner, registry and presets. A change to either can alter what every
+              # case-study pipeline does, and neither was listed - so a change to the module
+              # they all import fanned out zero case-study jobs and reported a green PR that had
+              # exercised none of them. The per-case-study filters below cover a study's own
+              # notebooks; nothing covered the code underneath all of them.
+              - 'case_studies/research/**'
+              - 'case_studies/utils/**'
               - 'data/**'
               - 'pyproject.toml'
               - '.github/workflows/test.yml'
--- a/case_studies/research/holdout.py
+++ b/case_studies/research/holdout.py
@@ -110,6 +110,16 @@ class StrategyReplay:
     def run(self, prediction: PredictionResult) -> BacktestResult:
         if prediction.study != self.study:
             raise ValueError("holdout prediction belongs to another study")
+        # This method loads holdout prices unconditionally, and `Strategy` exempts validation
+        # predictions from the holdout contract - correctly, because validation runs are ranked
+        # against each other and re-run freely. Handed a validation prediction, the two rules
+        # combine into a backtest of validation scores against holdout prices that no check
+        # would catch. The split is therefore refused here, where the prices are chosen.
+        split = prediction.registry_record()["split"]
+        if split != "holdout":
+            raise ValueError(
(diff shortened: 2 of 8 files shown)


stefan-jansen:
Superseded: the research-lock layer this teaches an expectation to is being deleted outright, so the expectation is moot. The second commit c1ca2183 (case-study matrix fans out on case_studies/research/** and case_studies/utils/**) is being cherry-picked into the removal branch - that defect is independent of the holdout question and affects every shared-infrastructure PR. Branch left in place until that pick lands.

PR #709, opened 2026-09-02
ch14 insights: compare aligned targets at float32 resolution

stefan-jansen:
`14_latent_factors/09_case_study_insights` raised `ValueError: Duplicate validation keys for sp500_equity_option_analytics latent` and is the current `ch14-15` failure on `main`.

The keys were not duplicates. Two latent configurations reached validation with metrics that agreed to float64 noise, so the comparison that de-duplicates them saw one key twice. Aligning the targets and comparing at float32 resolution separates them, and the notebook is re-executed against the nine production registries so its render matches the code.

- `14_latent_factors/09_case_study_insights.py` - compare aligned targets at float32
- `14_latent_factors/09_case_study_insights.ipynb` - re-executed render

Fixes the `ch14-15` job.


christy108:
Hi Stefan,

I want to unsubscribe, there are too many messages.

Thank you.
Christy

Issue #493, opened 2026-08-08
cross_sectional_ic_mean silently returns NaN: drop_nulls does not drop NaN in polars


stefan-jansen:
One line, copy-pasted into 19 notebooks across chapters 11, 12 and 13. It makes a notebook report `nan` for its headline metric, or silently select an arbitrary model, with no error.

## The bug

```python
ic_clean = ic_per_date.drop_nulls("ic")
return float(ic_clean["ic"].mean()) if ic_clean.height else float("nan")
```

In polars `null` and `NaN` are different things. `drop_nulls` removes nulls and leaves `NaN` in place, and one `NaN` poisons the mean:

```
drop_nulls leaves NaN in place: [0.1, nan, 0.3]
mean of that                 : nan
drop_nans then drop_nulls    : 0.2
```

## How it fires

`cross_sectional_ic_series` returns `NaN`, not null, for a date where Spearman is undefined because the predictions tie. A weak model config ties on some dates. Measured on `12_gradient_boosting/07_hpo_comparison` fold 0: **231 validation dates, 0 nulls, 52 NaN**. So `drop_nulls` removed nothing and the mean came out `nan`.

Every one of the 19 call sites writes `.drop_nulls("ic")`, so 19 independent authors all expected undefined dates to arrive as null. Worth deciding separately whether `cross_sectional_ic_series` should return null rather than `NaN` for a degenerate date, which would fix all 19 at the source.

It stays invisible until some config is weak enough to tie, which is why it has never been seen. A 0.45% change in training rows was enough to trigger it here.

## Fix

```python
ic_clean = ic_per_date.drop_nans("ic").drop_nulls("ic")
```

## Already fixed on `fix/openmp-import-order` (PR #491)

`12_gradient_boosting/{03_dl_vs_gbm,07_hpo_comparison,08_shap_analysis,11_conformal_gbm}` — the files that PR already re-executes.

## Remaining, needs one re-execution each

```
11_ml_pipeline/01_ols_inference.py
11_ml_pipeline/04_nested_cv_hpo.py
11_ml_pipeline/05_shap_analysis.py
12_gradient_boosting/02_gbm_comparison.py
12_gradient_boosting/04_optuna_tuning.py
12_gradient_boosting/05_cross_library_hpo.py
12_gradient_boosting/06_optuna_multi_asset.py
12_gradient_boosting/09_xai_limitations.py
13_dl_time_series/04_transformers.py
[...]

PR #713, opened 2026-09-02
Chapter 11: make linear SHAP actually exact, and stop the panel boots…

vkovshov:
…trap overstating its precision (#585)

* ch11 01_ols_inference: count autocorrelation lags on sessions, not rows

The residual autocorrelation diagnostic paired residuals by row position. A symbol's rows skip trading sessions wherever a feature could not be computed or the ETF was not quoting, so a pair labelled "lag 1" could be separated by any number of sessions. Three symbols in the training window resume after a break of a year or more, the widest being 1,764 sessions, and Durbin-Watson was computed across those breaks as if they were consecutive days.

Each residual is now placed at the number of the session it belongs to, and a pair counts towards a lag only when its two sessions are exactly that far apart. Durbin-Watson counts consecutive-session pairs only. Symbols whose measurable pairs at a lag fall below 60 are reported as unmeasured rather than silently averaged in. The direction of the finding is unchanged, as the diagnosis predicted: gaps attenuate a correlation rather than manufacture one.

The mean out-of-sample IC dropped nulls but not NaN. Polars treats them as different values, and cross_sectional_ic_series returns NaN for a date where every prediction ties, so one such date would have made the reported IC NaN. It now drops both. No date in this fold ties, so no value moved from it.

Numbers moved, cause by cause:

  Upstream: the ETF feature and label artifacts were regenerated on 2026-08-10,
  after this notebook was last executed on 2026-07-31. Every regression figure
  moves with them.
    rows              393,985 -> 402,446
    training obs      226,959 -> 228,787
    in-sample R2       0.0854 -> 0.0872
    out-of-sample R2  -0.1329 -> -0.1070
    IC mean           -0.0084 -> -0.0060
    IC t-stat HAC      -0.22  -> -0.15   (p 0.830 -> 0.882)
    IC t-stat naive    -0.60  -> -0.42
    significant at 5%: OLS 45 -> 44, date-clustered 28 -> 31,
                       two-way 21 -> 24, Driscoll-Kraay 5 -> 5

  Session-adjacency fix, on top of the upstream move:
    median within-symbol Durbin-Watson  0.1159 -> 0.1147
    median autocorrelation, lag 1       0.9414 -> 0.9425
    median autocorrelation, lag 5       0.7338 -> 0.7353
    median autocorrelation, lag 21     -0.0511 -> -0.0541

Also in this pass:

- The panel the analysis runs on is now shown before the diagnostics: the feature families that make the collinearity finding what it is, and a per-symbol coverage timeline that makes the unbalanced history visible (T13). The timeline is what the session-grid argument then points back to.

- The five-estimator standard error comparison was prose plus a nine-column table of ten features. It is now drawn: one box per estimator, showing its standard errors relative to the OLS ones over all features (T4, T5).

- Two prose sentences asserted results. One named the feature count as a literal and the other stated this run's out-of-sample R2 sign; both now say what to read and where, and the values stay in the cell output (N2, N3).

- Every figure carries alt text through utils.style.show_with_alt. The four images previously rendered as "No description has been provided for this image".

- The three subsampling draws used three literal seeds; they now use SEED (M4).

* ch11 01_ols_inference: plain dashes in prose

The reviewed notebooks in the corpus carry no em dash; this one carried 18. House style is the plain dash.

* ch11 02_regularization_paths: age recency weights in sessions, and share one IC helper

Two defects, one shared helper, and a presentation pass.

THE RECENCY WEIGHT WAS PER ROW, NOT PER SESSION. The sample-weighting cell demonstrates exponential decay, w_t = exp(-lambda * (T - t)), and computed the age from the row's position in the training array. This is a panel: the fold carries 165,105 rows over 2,240 sessions, about 74 rows per session, so lambda decayed per row and the weight fell roughly 74 times faster than the prose claimed. Two symbols quoting on the same day were also given different weights, which no recency argument justifies. The effective sample size was 1,001 rows, about the last ten sessions, so the cell was not demonstrating recency weighting: it was demonstrating a fit on almost no data.

Each row is now aged by the session it belongs to, and the decay is declared as a half-life in sessions rather than a bare lambda, because a half-life is the form a reader can act on.

  effective sample size  1,001 rows (1% of the fold) -> 31,206 rows (19%)
  IC weighted                          -0.1323 -> -0.1023

Weighting still costs IC on this fold. That is now a fact about the fold rather than an artefact of the decay rate, and the prose says what to read off the cell instead of asserting the sign.

THE MEAN IC DROPPED NULLS BUT NOT NaN (#493). A date whose predictions all tie has an undefined Spearman coefficient. ml4t-diagnostic 0.1.2 reports that as null and 0.1.1 - the version this environment resolves - as NaN, which drop_nulls does not remove, and one survivor turns the mean of the series into NaN. A fully penalized model predicts one constant and ties every date, which is why the LASSO summary already carried a filter for NaN rows.

Measured on this notebook: 397 validation dates, 0 undefined. So the fix moves no number here. It is a guard, and the fixture in tests/ is what keeps it one.

ONE IMPLEMENTATION INSTEAD OF THIRTEEN. utils.modeling.cross_sectional_ic_mean now holds the numpy-array adapter over ml4t.diagnostic's cross_sectional_ic_series, which thirteen notebooks across chapters 11, 12 and 13 had each copied. This notebook is the first to call it; the rest follow as they are fixed. The library computes the correlation, this only adapts four aligned arrays to the two frames it wants. Its import is function-local because ml4t.diagnostic pulls scikit-learn's OpenMP runtime up transitively and chapter 12 must load a gradient-boosting library first.

tests/test_cross_sectional_ic_mean.py pins the behaviour: a tied date is dropped rather than propagated, an all-tied series returns NaN rather than 0.0, and min_obs excludes a thin cross-section. One test first establishes that the library really does leave a tied date undefined, so the others cannot pass for the wrong reason if that changes.

NUMBERS THAT MOVED FOR THE DATA, NOT FOR EITHER FIX. The ETF feature and label artifacts were regenerated on 2026-08-10, after this notebook last ran.

  rows                   393,985 -> 402,446
  training rows, last fold  157,172 -> 165,105
  OLS mean IC              0.0284 -> 0.0231
  Ridge IC at its alpha    0.0473 -> 0.0481
  LASSO IC at its alpha    0.0448 -> 0.0480
  IC unweighted           -0.0009 -> 0.0160
  mean rank correlation    0.7596 -> 0.7645
  best ICIR                  1.30 -> 1.36
The selected alphas did not move.

PRESENTATION

- The alpha for each penalized model is chosen on the same validation folds the score is then reported from. The notebook now says so, and says that this is what 04_nested_cv_hpo exists to separate (C2).

- Numbers are out of the prose. Two ICs, three loss-comparison ICs, a feature count and a survivor count were literals in markdown from an older data vintage; the text now says what to read and the values stay in the cell output (N1, N2). A figure title interpolating the selected alpha is now a claim (N4).

- Three figures were duplicates: the Ridge IC panel repeated the later IC/ICIR figure, and a combined three-panel repeated the LASSO path and the model comparison. The comparison table is replaced by the bar chart carrying its fold-to-fold spread, and the Ridge coefficient path now stands on its own, which is what makes the contrast with the LASSO path readable (T4, T5).

- Every figure carries alt text, and the eleven em dashes are plain dashes.

The joblib cache was cleared before this execution, so the committed outputs are derived from the artifacts and not from a cache hit (C9).

* ch11 05_shap_analysis: HAC t-stat on the IC, and takeaways that do not depend on one fold

The IC t-statistic divided by the standard error of independent observations. Each date's IC is measured against a 21-day forward return, so consecutive daily ICs share twenty of twenty-one days of outcome window and are correlated by construction. It now uses compute_ic_hac_stats with the lag set to the label horizon, the same call 01_ols_inference makes, and prints the naive statistic beside it so the size of the difference is visible.

  IC t-stat  1.17 naive -> 0.50 HAC (naive on the new vintage is 1.38)

The mean IC also dropped nulls but not NaN, so a date whose predictions all tied would have made it NaN. It now drops both.

Numbers moved with the ETF artifacts, regenerated 2026-08-10:

  rows                    393,985 -> 402,446
  test fold                22,170 -> 22,184 samples over the same 252 dates
  IC                       0.0183 -> 0.0216
  SHAP vs closed form       0.9928 -> 0.9808 correlation
  wrong at high conviction     58% -> 58%
  mean max-feature share     19.7% -> 19.1%

PRESENTATION

- The four key takeaways named specific features, a wrong-call percentage, a correlation and a concentration share, all from the 2026-07 vintage and all wrong by the time anyone read them. They now state what the reader should do with each diagnostic: what mean absolute SHAP does and does not establish, why the right-versus-wrong split is what turns attribution into diagnosis, and how cross-fold and within-fold instability differ (T3, N2).

- Three more prose passages carried run values, including one that framed the high-conviction hit rate. Removed; the cell output above each carries them.

- The alpha was described as "a moderate default". 02_regularization_paths sweeps this same panel and finds the useful range four to six orders of magnitude higher, so the description was not true of the sibling notebook it implicitly cites. The prose now says the penalty is deliberately light, what that buys for reading a decomposition, and that it is not the penalty that notebook selected (N3).

- Nine figures had slogan titles and no alt text. Titles are claims now, and every figure goes through utils.style.show_with_alt.

- The npz sidecar said "downstream consumers can load these". The consumer is the book repo's Figure 11.3 generator, which is now named (C13).

- Twenty-six em dashes are plain dashes.

* ch11 05/02: an exact SHAP background, a panel-aware bootstrap, and a recency bet not confounded with its penalty

05_shap_analysis, the blocking one. The prose said linear SHAP here involves 'no approximation, no sampling' while shap.maskers.Independent(X_train) took the library default max_samples=100, so the background expectation was a 100-row subsample of 165,105 training rows and the closed-form check printed correlation 0.980770 with a max difference of 1.18e-02 directly above a line claiming the attributions match. The masker now takes the full training set.

old -> new -> cause: correlation 0.980770 -> 1.000000, max |difference| 1.18e-02 -> 0.00e+00, because the background is now the mean the closed form uses rather than a sample of it. The section's claim is exact as written, which is the point it exists to teach.

05, the bootstrap. It resampled 22,184 test rows IID, but the fold is a panel: about a hundred ETFs quote on each date and adjacent dates share overlapping 21-session feature windows, so the intervals were far too narrow. It now resamples dates with their full cross-section.

05, the ranking rule. 'A feature whose interval excludes another feature's point estimate is genuinely more important' compares two different quantities, and the two move together across replicates - a session that lifts one lifts the other - so it ignores the correlation that decides the question. A new cell bootstraps the paired difference on the same redrawn sessions and claims a ranking only where that interval excludes zero. vol_126d outranks 9 of 9 at 95%.

05, two prose defects. Ridge does not structurally bound any single feature's attribution share; 19.1% mean and 31.2% at the 95th are properties of this fit, not of the penalty, and the text now reports them as such. And a plt.figure(figsize=(8,5)) that shap.plots.scatter never drew into, emitting a stray '<Figure size 800x500 with 0 Axes>', is deleted.

02_regularization_paths. The weighted and unweighted Ridge both took alpha=best_ridge_alpha, but sklearn scales rows by sqrt(w) without renormalizing, so with a mean weight near 0.19 the penalty was roughly five times stronger relative to the data term in the weighted fit. The printed IC drop therefore confounded the recency bet with a large change in regularization, while the prose attributed all of it to the bet. Weights are normalized to mean one before fitting.

old -> new -> cause: IC weighted -0.1023 -> -0.0867, because the weighted fit is no longer carrying five times the penalty. Unweighted is unchanged at 0.0160 and the effective sample size is still 19% of the fold, both by construction. The sign does not move, so the section's reading stands; the size of the gap now measures what the section says it measures.

02, prose. Two references to 'the table' in the Model Comparison section, where the same earlier commit had replaced the frame with a bar chart.

Both notebooks re-executed and stamped.

* ch11 02/05: effective sample size that survives rescaling, and prose that matches the run

02, the statistic. n_eff was weights.sum(), taken before the fit's weights are rescaled to mean one. Sum-of-weights is scale-dependent, so it answered for a vector the code discards one line later; over the weights actually used it is exactly n_train, which is 100% of the fold and says nothing. It is now Kish's (sum w)^2 / sum(w^2), which is unchanged by the rescaling, and the markdown states that definition and why the plain sum will not do.

old -> new -> cause: effective sample size 31,206 rows (19% of the fold) ->
60,405 (37%), because the plain sum understates how much of the sample a half-life of 252 sessions actually retains. The section's point is unchanged - the fit still gives up most of its rows - but the size of what it gives up was overstated by nearly a factor of two.

That number came out of the joblib cache on the first two re-runs, not out of the new code: RETRAIN is False by default and NEED_TRAINING is false whenever cv_results.joblib exists, so the notebook printed 19% while the cell above it computed something else. The cache is gitignored, so a clean clone always trains and no reader would have seen it, but the render nearly went out with a number its own code no longer produces. Retrained from an empty cache.

02, the cache contract. The load path reads n_sessions and half_life, which this branch introduced. A cv_results.joblib written by the previous version has sample_weight_results without them, so a reader who had run the notebook before and reran with RETRAIN=False got KeyError: 'n_sessions'. A results dict missing any expected key is now treated as absent and the reader gets the RETRAIN message.

05, the lead-in. It told the reader SHAP values 'should closely match' the coefficient attribution and that the library's background handling 'may introduce tiny numerical differences, but the correspondence is near-perfect'. With the full-training-set background the cell prints correlation 1.000000 and max difference 0.00e+00, and Key Takeaway 1 says 'no approximation and no sampling'. The paragraph promised a discrepancy the notebook no longer produces and that the masker change was made to remove. It now states the identity holds exactly, and keeps the caveat as the condition a subsampled background would introduce.

Both notebooks re-executed and stamped.

* ch11 05/02: the other masker, a cache marker, and a render that ran off the cache

Four things, all consequences of the two commits before this one.

05 constructs shap.maskers.Independent twice. I passed max_samples=len(X_train) in the NEED_TRAINING block and left the stale-artifact recompute branch on the library default of 100 rows, so whenever that branch fires the verify cell would print about 0.98 correlation and a max difference near 1e-2 directly under a paragraph asserting the identity holds exactly. That branch also writes shap_arrays.npz, which is the source of Figure 11.3. Both blocks now use the full training set.

02's cache guard did not detect the staleness the commit that added it created. It required six keys to be present, and a cache written by the immediately preceding commit has all six with n_eff as the plain weights.sum() - it passes, and the reader is shown 19% of the fold where the notebook computes 37%, with no prompt to retrain. sample_weight_results now records n_eff_kind='kish' and the guard requires it, so any cache predating the formula falls through to the RETRAIN message.

05 carried two statements about the same measurement that disagreed: a comment saying a 100-row background leaves the check off in the third decimal, prose saying the second. Measured max difference under the default masker is 0.0118, so the comment was the stale one.

And the render itself. 05's previous execution ran off its joblib cache, so the committed notebook had three empty code cells where the published page prints 'Ridge trained on 165,105 samples, predicting 22,184.', 'SHAP values computed: (22184, 57)' and 'Computed SHAP stability across 8 folds.' A reader would have seen the training and SHAP-computation steps produce nothing. This is the cache-hit hollowing hazard: re-running an already-cached notebook short-circuits the work, emits no output, and re-stamps production: True regardless. Both notebooks are re-executed here from cleared caches, so the training path is the one shown.

Verified in the committed .ipynb: correlation 1.000000, mean and max difference 0.00e+00, effective sample size 60,405 rows (37%), IC unweighted 0.0160 and weighted -0.0867.

* ch11 05: the fold-stability loops had their own background, and the cache did not record which one it used

The notebook constructs shap.maskers.Independent four times. The two previous commits fixed the main explainer and the stale-artifact recompute branch; the two fold-stability loops still took the library default of 100 background rows. The consequence is checkable by a reader: fold 8 of that chart is the same split, model and scaling as the importance table, so its point for a feature should equal that feature's mean_abs_shap, and on two different backgrounds it did not. Under a percent, so the chart's message held, but it is a visible disagreement in the section that has just argued the full background is what makes the attributions exact. Both loops now use it.

The SHAP cache got no counterpart to the n_eff_kind marker added to 02_regularization_paths. shap_results.joblib is gitignored and machine-local, and both staleness tests compare fold count and row count only, so a cache written before the background change passes: the reader loads 100-row-background shap_values while coef_attribution is computed from the cached full X_train mean, and the verify cell prints 0.980770 and 1.18e-02 under prose saying the two agree to float precision. The cache now records background='full' and a cache without it is treated as stale.

Re-executed from a cleared cache. Verified in the committed .ipynb: 'Ridge trained on 165,105 samples', 'Computed SHAP stability across 8 folds', correlation 1.000000, mean and max difference 0.00e+00.

* ch11 05: the staleness guard read a dict the cell above had deleted

_cache_background_stale read _cached, and the loading cell ends with del _cached. On the cached path the left operand of the  is True, so the right one evaluates and raises NameError. That is exactly the path the guard exists to protect: a reader who has run the notebook once and reruns with RETRAIN=False aborted at that cell instead of recomputing. It was invisible in the committed outputs because I executed from a cleared cache, where NEED_TRAINING is True and the branch never runs.

The marker is now read in the load cell, before the dict goes.

Verified by running the three paths rather than reasoning about them:

  cleared cache     -> trains, 0 error outputs, prints 'Ridge trained on
                       165,105 samples' and correlation 1.000000
  warm cache,
  background=full   -> 0 error outputs, no recompute, no stale message
  pre-fix cache,
  background absent -> 'Cached SHAP artifacts are stale (fold count, data
                       vintage or background changed); recomputing full
                       artifacts', then correlation 1.000000

The third is the case that mattered: before this change that cache loaded silently and the verify cell printed 0.980770 with a max difference of 1.18e-02 under prose saying the two agree to float precision.

Committed render is the cleared-cache run, so the page shows the training path and not a maintainer-facing staleness message.

* Strip the machine-specific paths the ch11 renders carried into their outputs

Same leak as the ch16-17 branch: 02_regularization_paths committed two absolute paths from this workstation and 05_shap_analysis one. The check lives in the required guards job, so this was blocking the merge. Rewritten by .github/scripts/sanitize_notebook_paths.py; all 14 hygiene tests pass.


stefan-jansen:
Closing: the content of this PR is already on `main`.

The Chapter 11 commit here (`d991c156`, "make linear SHAP actually exact, and stop the panel bootstrap overstating its precision") landed via #585. Diffing this branch against current `main` leaves only `uv.lock` - 20 lines of lockfile drift picked up by the `Merge branch 'main' into master` commit, nothing else:

```
$ git diff --stat origin/main...pr713
 uv.lock | 24 ++++++++++++++++++++----
 1 file changed, 20 insertions(+), 4 deletions(-)
```

Thanks for the contribution - the substance is shipped. Nothing here is being rejected; there is simply nothing left to merge.

PR #719, opened 2026-09-02
Filter the top-k concentration grid at the long-short ceiling

stefan-jansen:
Closes ml4t/agent-workspace#401.

`get_top_k_values_for` filtered only `k >= n_assets`. A long-short strategy takes `k` names on each side, so its ceiling is `n_assets // 2`: `signals.py:217` clamps `eff_k` to that, while `top_k` is identity-bearing through `plan_backtests(signal=...)`. A `k` in the half-open interval `(n_assets // 2, n_assets)` therefore passed the filter, registered under the declared `k`, and ran at the clamped one - and two such values would register distinct identities over a byte-identical weight series.

`get_entry_schemes_for` has excluded that interval since it took a `long_short` argument. Its own comment says it excludes `k >= n_assets` "to match get_top_k_values_for", so the two were written to agree and did not.

## Where the flag comes from

`long_short` defaults to the case study's declared selection mode - `mapping.position_state_space`, via `get_backtest_config`, which is the same value every caller already passes to `get_entry_schemes_for` as `bt_config.long_short`. Not `account.allow_short_selling`: that is an execution permission and the two come apart. `sp500_options` sets the permission because it sells straddles, while its state space is `short_straddle_hedged` - one-sided, with no cross-sectional short sleeve - and `12_backtest` passes `long_short: False` to match. The first commit here used the account flag; the review caught it and the second commit corrects it.

Defaulting rather than requiring the argument keeps this to one file. All twelve call sites pass only `(case_study, label, n_assets)` and none has a resolved strategy spec at that point, so a required argument would be ten notebook edits, ten re-executions for provenance, and ten chances to pass the wrong value.

## No registered identity moves

Measured across all nine case studies at their declared universes, the returned grid changes for `crypto_perps_funding` alone:

```
crypto_perps_funding  fwd_ret_8h     [3, 5, 10] -> [3, 5]
crypto_perps_funding  fwd_ret_24h    [3, 5, 10] -> [3, 5]
crypto_perps_funding  fwd_dir_8h     [3, 5, 10] -> [3, 5]
crypto_perps_funding  fwd_dir_8h_3c  [3, 5, 10] -> [3, 5]
```

and `crypto_perps_funding` reaches its concentrations through `get_entry_schemes_for`, which already excluded k=10. Its registry holds 2,852 backtests at `top_k` 3 and 5 and nothing at 10, so nothing registered has to be recomputed. Every other case study is unchanged: `cme_futures` passes 30 products against a grid topping out at 10, `fx_pairs` declares k=10 on exactly 20 pairs, and the rest have ceilings orders of magnitude above their grids.

## Tests

`test_the_two_grid_helpers_agree_on_every_declared_case_study` pins the agreement across all nine case studies at their declared universes. It fails on `crypto_perps_funding` without the change. `test_the_ceiling_follows_the_selection_mode_not_the_account_permission` asserts `sp500_options` at 30 assets, where the selection mode gives `[5, 10, 20]` and the account permission would give `[5, 10]` - the point where the two sources disagree.

```
48 passed, 10 skipped   (test_sweep_config_seam, test_unfinished_sweep_plans, test_crypto_cost_sweep_contract)
```

https://claude.ai/code/session_01NSbpWxbSRQp8AsVDJXcpPi


stefan-jansen:
Holding this as a draft: it turns `cs-nasdaq100_microstructure` red, and the reason is worth a decision rather than a workaround.

## What fails

`cs-nasdaq100_microstructure` passes on `main` (runs 33639112144, 33603511524) and fails here. Reproduced locally:

```
long_short: True
n_assets=6:   ValueError: top_k_grid['fwd_ret_5m'] = [5, 10, 20] is empty after filtering
n_assets=20:  [5, 10]
n_assets=115: [5, 10, 20]
```

`tests/overrides.yaml` gives `nasdaq100_microstructure/15_portfolio_management` `MAX_SYMBOLS: 6`, because the fixture holds 6 nasdaq symbols. On 6 assets long-short, the smallest declared k of 5 needs 10 distinct names, so every member of the grid is unrealizable and the existing "empty after filtering" guard fires. `14_backtest` is unaffected: it takes `TOP_K: 2` from the same overrides, and its `get_entry_schemes_for` call returns 38 slot schemes and zero `equal_weight_top_k` at that width.

## Why this is the issue rather than a side effect

The override entry already documents the situation, in its own words:

> 6 is the whole fixture universe and is still not enough to realize top_k=5 long-short - signals.py caps each side at `n_assets // 2`, so these rows trade 3 names per side under a top_k=5 label.

That is ml4t/agent-workspace#401 exactly: a row registered under a `top_k` the engine did not run. It is live in CI today, accepted in a comment, and this change converts it from a silent false identity into a refusal. The refusal is the behaviour I would want; making it quiet again would defeat the point of the change.

## The decision that is not mine

`nasdaq100_microstructure` is mid-rebuild, so I am not touching its notebooks or its fixture. Three ways out, in the order I would rank them:

1. Widen the nasdaq fixture past 10 symbols so `top_k=5` long-short is realizable. Fixes the CI row's identity as well as the red.
2. Give `15_portfolio_management` the `TOP_K` parameter `14_backtest` already has and set it to 2 in overrides. Small, and makes CI declare a k it can realize.
3. Declare a smaller k in `backtest.sweep.top_k_grid` for the case study. Changes production, so not this.

Everything else here is green, including all four long-short case studies whose grids this touches: `cs-crypto_perps_funding`, `cs-cme_futures`, `cs-fx_pairs`, `cs-sp500_options`, plus `cs-us_firm_characteristics` and the five required checks.

PR #724, opened 2026-09-02
sweep_config: a long-short top_k above half the panel is not feasible

stefan-jansen:
`signals.py:217` clamps each side of a long-short selection to `n_assets // 2`. A `top_k` above half the panel therefore executes with fewer names than the spec it registers under, and two declared `k` values collapse to the same portfolio while registering as distinct `backtest_runs` rows.

`get_entry_schemes_for` already had the right test (`long_short and 2 * k > n_assets`). Two siblings in the same file did not:

- `get_signal_nasdaq100_schemes_for`'s `eq_w_topk` branch tested only `top_k >= n_assets`, while setting `long_short` from its own direction axis.
- `get_top_k_values_for` tested only `k < n_assets` and had no way to know the direction. Rather than add a parameter that twelve call sites could forget, it now derives it from the case study's own `mapping.position_state_space` via a new `is_long_short()`, using the same token rule as `backtest_loaders.load_backtest_config` - so sp500_options' one-sided `short_straddle_hedged` correctly does not count.

## No production value moves and no registered row is invalidated

Verified against all nine case studies at production width; every declared grid was already feasible.

| case study | long-short | n | offered |
|---|---|---|---|
| us_firm_characteristics | yes | 2500 | 5, 10, 20, 50 |
| nasdaq100_microstructure | yes | 100 | 5, 10, 20 |
| fx_pairs | yes | 20 | 5, 10 |
| cme_futures | yes | 30 | 5, 10 |
| crypto_perps_funding | yes | 19 | 3, 5 |
| us_equities_panel | yes | 2000 | 20, 50 |
| etfs | no | 60 | 5, 10, 20 |
| sp500_equity_option_analytics | no | 200 | 5, 10, 20 |

crypto's `[3, 5]` is exactly what its registry holds, as are cme's `[5, 10]` and fx's `[5, 10]`.

The only value that moves is at CI fixture width, which is where this surfaced: nasdaq100 at 12 symbols offered `top_k=10`, executed it with six a side, and registered it as ten. It now offers `[5]`.

The empty-grid error also stopped claiming the wrong cause - for a long-short case study it now says the k needs more than half the panel a side, and names the universe size that would admit it.

## Tests

`TestLongShortHalfPanel` in `tests/test_sweep_config_seam.py`: the long-short drop, the long-only non-drop, the corrected error message, and a parametrized check that every k still offered at production width is fillable. 109 tests pass across the sweep surface.

No notebook changed, so nothing needs re-running.

### DIFF
--- a/case_studies/utils/sweep_config.py
+++ b/case_studies/utils/sweep_config.py
@@ -22,6 +22,8 @@ Usage:
 
 from __future__ import annotations
 
+import re
+
 import polars as pl
 import yaml
 
@@ -612,7 +614,12 @@ def get_signal_nasdaq100_schemes_for(
             for direction in directions:
                 ls = direction == "long_short"
                 for top_k in top_k_grid:
-                    if top_k >= n_assets:
+                    # `signals.py` clamps each side of a long-short selection to
+                    # `n_assets // 2`, so a k above half the panel executes with
+                    # fewer names than it registers under. Two declared k values
+                    # then collapse to the same portfolio while registering as
+                    # distinct rows. Same test as `get_entry_schemes_for`.
+                    if top_k >= n_assets or (ls and 2 * top_k > n_assets):
                         continue
                     dtag = "ls" if ls else direction[0]
                     schemes.append(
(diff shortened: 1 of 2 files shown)


stefan-jansen:
Duplicate of #719, which does the same fix better - it delegates to `get_backtest_config` instead of re-implementing the `position_state_space` token rule, takes an explicit `long_short` override, and writes an error message that names how many names the sweep actually needs. Closing this in its favour.

One piece here is not in #719 and I am moving it there rather than losing it: `get_signal_nasdaq100_schemes_for`'s `eq_w_topk` branch has the same defect. It tests only `top_k >= n_assets` while setting `long_short` from its own `direction` axis, so at 12 symbols it still offers `ewtopk_ls_k10` and executes it with six a side. That is the second of the two places the nasdaq lane hit.

Issue #187, opened 2021-08-03
chapter14: 06_sentiment_analysis_yelp.ipynb


Eeonum:
The yelp data url needs to be updated to:

`https://www.yelp.com/dataset`

stefan-jansen:
The [yelp data download notebook](https://github.com/stefan-jansen/machine-learning-for-trading/blob/master/data/create_yelp_review_data.ipynb) uses the above url.

PR #736, opened 2026-09-03
nasdaq100_microstructure: one-minute price grid, decisions on each label's horizon

stefan-jansen:
Settles [#187](https://github.com/ml4t/agent-workspace/issues/187) and [#1005](https://github.com/ml4t/agent-workspace/issues/1005).

The labels spanned `[t+1min, t+15min]` on the minute grid while the engine ran a fifteen-minute clock and held `[t+15min, t+30min]`. Two intervals of the same length sharing no minutes, so the model's skill could not reach the P&L by construction.

## The design

Two grids, separated rather than collapsed onto one:

| | |
|---|---|
| **price grid** | one minute - what a fill and a stop are measured on |
| **decision cadence** | the model's own horizon, per label: 5, 15 or 60 minutes |
| **the minutes between** | position monitoring, which is what makes a stop or take-profit expressible at all |

It also aligns the fill with the label for free: `execution_mode: next_bar` on a minute feed enters at `t+1` minute, which is where the rewritten label prices its entry leg.

## What was actually wrong

`resolve_rebalance_timestamps` returned an intraday panel's timestamps **unchanged**, on the comment that the data was already at the right granularity. That is a precondition and nothing checked it. This panel carries every minute - `03_financial_features` emits 16.9M rows and thins to the decision grid for two figures only - so a declared fifteen-minute cadence produced a decision **every minute**.

The resolver now constructs the schedule from the declared cadence, anchored on the clock. That is the grid `03_financial_features` already states as `minute % DECISION_MINUTES == 0`, so a fifteen-minute cadence decides on the quarter hour in every session, whatever time the panel starts and however early the session closes.

With the cadence carrying the horizon, every `rebalance_step` is 1, including `fwd_ret_60m`, whose `ceil(60/15) = 4` the per-label cadence now expresses directly.

## The step is part of the identity it decides (#1005)

The step chose which slots were traded and therefore every metric recorded, but two runs of one configuration at different steps hashed identically and the second was skipped as already registered. `build_backtest_spec` now emits the resolved step beside the cadence, `run_backtest` stamps it where both the case study and the label are always in scope, and all three execution paths read it from the frozen spec rather than from mutable `setup.yaml`. Emitted only where a step is declared, so a case study that declares none produces byte-identical specs.

## The completed case studies do not move

- Daily and coarser cadences name no interval and never reach the constructed branch, so `etfs`, `cme_futures`, `fx_pairs`, `us_firm_characteristics` and `sp500_equity_option_analytics` resolve exactly as before.
- `crypto_perps_funding` does take it, and a test asserts its 8-hourly panel under its own cadence resolves to itself. Its `fwd_ret_24h` steps 3 slots, and folding the step into the interval makes that schedule *stop* depending on outages: under `gather_every` a missing funding time re-phased every decision after it permanently.
- `nasdaq100_microstructure` holds zero backtest rows, so no registered result changes.

## Verification

`tests/test_intraday_decision_schedule.py` (9) and `tests/test_rebalance_step_identity.py` (3). Each pins the reading the *old* code produced as well as the new one, so a revert fails rather than passing quietly.

Affected suites go 118 to 289 passed on the same four pre-existing failures - three missing local market data, one `test_model_registry` case - all four failing identically with these changes stashed.

## Not in this PR

The case study has not run. This makes the run correct to start; it is not the run.


christy108:
I WANT TO UNSUBSCRIBE I FED UP DO THESE MESSAGES. IVE EMAILED YOU MANY TIMES. IVE ALSO UNFOLLOWED IN GITHUB AND "UNSUBSCRIBED" BUT STILL GET THESE MESSAGES

PR #759, opened 2026-09-04
An uncovered training row is no longer fitted as the feature mean

stefan-jansen:
`model_based.parquet` carries one fold geometry, written by the primary label. A label whose window is wider has rows the artifact never covered, and every join that attaches it is a LEFT join, so those rows survived with null temporal features.

Nothing downstream removed them. The filter after the join tests the **label** for null, not the features, and `sequence_dataset.py:217` calls `np.nan_to_num(..., nan=0.0)` — which, after the per-feature normalization, is the feature's **mean**. So an uncovered row was not fitted as missing. It was fitted as an average observation, and no error, warning or metric showed it.

`validate_temporal_alignment` already declined to score these rows: it drops a leading warm-up prefix from the coverage denominator (`utils/modeling.py:1483-1490`) and scores the rest. That is why `etfs` passes at ~100% while **5,244 of 1,893,267 training rows** are null across 8 folds, in all **64 registered `fwd_ret_5d` fits**. The guard refusing to measure a window and the estimator being handed it anyway is the defect.

## One implementation, every consumer

`split_frames` is not the only join site, and fixing it alone would have missed most of the families. GBM, TabM, both shared modeling loops, the sequence builder and Darts all reach the artifact through `replace_temporal_columns`. The drop lives there behind `drop_uncovered=`, every training caller opts in, and `split_frames` uses the polars equivalent from the same module.

Coverage is **membership in the artifact** (the merge indicator), not a test on the feature values — a covered row may legitimately be all-null when temporal fitting skips an entity.

## Training only

Validation is left whole. A prediction set must cover every session its declared folds contain (`reference/CASE_STUDY_PIPELINE.md` §3) and the expected-key builders enforce that from the dataset, not the covered rows (`gbm.py:1668`). Trimming validation would convert a clean stop into a short prediction set the registry rejects later, naming the wrong cause. `drop_uncovered` defaults to `False`.

Darts overlays one frame spanning `train_start..val_end` and splits it afterwards, so it gets `drop_uncovered_through=train_end`: training half trimmed, validation half left for the guard.

## Bounded by what the guard excuses

The trim stops at `MAX_TEMPORAL_WARMUP_FRACTION`. Without that it would be strictly more permissive than the guard it serves — an artifact fold beginning years into a window would have its whole head removed in silence. `us_equities_panel/12_dl_weekly.py` is the live shape. An interior hole raises separately: windows are admitted by expected period number, so a trim only costs the windows spanning it, while a hole also costs the lookback windows either side.

## Identities move

`FOLD_PREPARATION_VERSION`, `SEQUENCE_PREPARATION_VERSION` and `TABM_RUNNER_VERSION` all go 1 → 2, with every pin re-pinned in the same commit. `linear` and `gbm` need no bump of their own — their source identities already carry `FOLD_PREPARATION_VERSION`. The pinned preparation digest does **not** move: that fixture carries no temporal artifact, which is the check that this is a no-op wherever there is nothing to cover.

## Measured scope

| case study | primary uncovered | variant excess-left | excess-right |
|---|---|---|---|
| `cme_futures` | 0/6 | 0/6 | 0/6 |
| `crypto_perps_funding` | 2/3 | 0/3 | 0/3 |
| `etfs` | 0/9 | **8/9** | **1/9** |
| `fx_pairs` | 8/9 | 0/9 | 0/9 |
| `sp500_equity_option_analytics` | 0/3 | 0/3 | 0/3 |
| `us_equities_panel` | 0/17 | 0/17 | 0/17 |

`fx_pairs` and `crypto` have real estimator warm-ups with no artifact rows to have, so the excuse is load-bearing and failing on any leading gap would break them on a condition that is correct there.

## What this does not do

It does not fix the artifact's geometry. `model_based.parquet` still carries the primary's folds, and making it per-label reopens accepted stage 01-05 work in six case studies — a scope decision, not a defect adjudication. Refs ml4t/agent-workspace#926.

922 related tests pass. The one `etfs/06_linear` failure in the sweep is the 231/247 validation raise, confirmed identical on a clean tree; it is ml4t/agent-workspace#1031 and predates this.

https://claude.ai/code/session_01B1EzmRdDAxsqksSBN9EsT3


stefan-jansen:
Closing without merging. The ruling on the contract behind this PR was to impute rather than trim: fix the sequence fill, do not drop the rows.

**Why the drop is the wrong half.** This PR removed uncovered training rows from the fit. Measuring what those rows actually are, split by cause on the production dataset, primary label, fold 0:

| case study | uncovered | skipped symbol | warm-up prefix | trailing | interior |
|---|---|---|---|---|---|
| etfs | 1,049 | 0 | **1,049** | 0 | 0 |
| sp500_equity_option_analytics | 3,929 | 3,432 | 99 | 398 | 0 |
| sp500_options | 2,657 | 2,356 | 50 | 251 | 0 |
| crypto_perps_funding | 590 | 590 | 0 | 0 | 0 |

The dominant cause is not fold geometry. It is symbols the temporal model skipped entirely, which `validate_temporal_fold_coverage` documents as legitimate and imputed downstream. etfs is the only case study whose uncovered rows are genuinely all warm-up prefix, and it is the one this PR was built on.

**The guard here is what was failing `cs-sp500_equity_option_analytics`.** Fold 0 of that fixture has 986 of 2465 training rows uncovered, 40.0%, past the 10% the coverage guard excuses. The drop refused a geometry that imputation handles. That is the measurement that makes the ruling reasonable rather than arbitrary: a rule that stops a case study over rows the design intends to impute is stricter than the contract it was written to serve.

**The defect this PR was really written against is already fixed on main**, by #767 (`174154ad`, "Sequence families: impute a missing feature at the training mean, not below its minimum"). In the sequence path `np.nan_to_num(..., nan=0.0)` ran *before* normalization, so a missing feature became a raw zero and was then standardized — on seoa's `garch_cond_vol` that lands at **-1.64 sd**, below the observed minimum, presenting a skipped symbol as the calmest stock in the panel; and because the zeros were present when the statistics were computed, they skewed every other symbol's normalization too. That is now imputed at the training mean.

**One piece of this PR is independently correct and moves to its own PR:** #767 changed what the sequence families are fitted on without moving `SEQUENCE_PREPARATION_VERSION`, which is still 1 on `main`. A registered version-1 run and a new run under the corrected fill would share an identity while having been fitted on different data.

Nothing else here travels. `FOLD_PREPARATION_VERSION` and `TABM_RUNNER_VERSION` stay at 1, the fold-cache key pin stays as it is, and `drop_uncovered` / `drop_uncovered_temporal_pl` are not needed.

Issue #926 stays open and is unaffected: it is being addressed case study by case study by the stage-04 refit-schedule work (#762, #764, #766, #769, #773).

PR #762, opened 2026-09-04
crypto_perps_funding 04: a refit schedule replaces per-fold fitting

stefan-jansen:
Depends on #760, which carries `walk_forward_feature`. **Code only - the notebook is committed cleared and has not been executed.**

## The notebook already knew

Its last "Known limitation" said this, and it was recorded rather than fixed:

> Within a fold, the parameters that produce a training row's feature were estimated from the whole training window, so an earlier training row carries a value fitted partly on returns that came after it. Validation rows do not.

That is the whole defect. A fitted feature has two channels through which the future reaches a value - the **conditioning set** and the **parameters** - and this notebook closed the first (`filtered_state_probs`, `.fix()`) while leaving the second open. Every assertion it ran checked the channel that was already clean.

## What changed

Both models are fitted on a schedule that runs the length of the history: burn in `MIN_TRAIN_BARS`, fit, emit until the next refit, re-estimate on everything up to that point, carry on. GARCH every 21 settlements, the regime chain every 63, both declared in `setup.yaml` under `model_based`. Parameters freeze at the holdout boundary via `freeze_after`, so the holdout gets values and none of them were informed by it.

**The `fold` column is gone.** Keys are `(symbol, timestamp)`. A settlement had a different value in every fold, which is what a fold-frozen fit produces and what made the artifact 1.91x the panel it describes. `load_modeling_dataset` takes its single-feature-set path - the one `nasdaq100_microstructure` already uses - and the notebook asserts `temporal_by_fold is None` so a surviving fold column cannot pass unnoticed.

Also moved: the seals check the schedule rather than a fold's training window; `garch_vol_zscore` is computed over `symbol` rather than `(fold, symbol)`; section D plots the fitted parameters as a time series rather than three points; the fold figure becomes a schedule figure; the artifact's metadata carries `estimation_schedule` where it carried `fold_geometry`; and a burn-in prefix now carries nulls on purpose.

## Verification

The second commit fixes three defects the pre-push review caught, all of them consequences of the rewrite: `arch` returns an ndarray when handed one so `.to_numpy()` would have raised on the first fit; NaN is a float in polars so the burn-in survived `drop_nulls` and the artifact's null counts would have read as zero on a column unavailable for 500 settlements; and section F still filtered on a `fold` column the write block no longer emits.

The GARCH callback was then run end to end against the pinned `arch` outside the notebook - 900 bars, burn-in 500, refit every 100, freeze at 800. Shape `(900, 2)`, exactly 500 leading NaN, 500 nulls after `nan_to_null`, and **the walk over a 700-bar truncation reproduces the first 700 rows of the full walk to 1e-9**. That last one is the property the whole change exists for, and it is what a fold-frozen fit fails.

`ruff check`, `ruff format`, `py_compile` clean; `jupytext --sync` run. Section D's quantitative reading is written when the notebook executes; the prose currently says what to read off each chart and asserts no value it has not computed.

## Cost of the re-run

All **139** of this case study's training runs refit, because `computation.feature_artifacts.model_based.sha256` is pinned in every training spec and is not in `_MIGRATABLE_FIELDS`. 678 prediction sets and 2,852 backtest rows sit behind them. Sequenced after the open issues close and against a fresh registry.


stefan-jansen:
Holding this as a draft: the emission path still reads `arch`'s result object, which is the defect this PR exists to close.

`frozen_garch_path` takes its `variance` from `np.asarray(fixed.conditional_volatility) ** 2`, where `fixed` is `arch_model(returns_prefix * 100, ...).fix(params)`, and only then advances one step by hand. `arch` derives the residuals, the backcast seeding sigma^2_0 and the variance bounds from whatever sample it is handed, so extending the prefix moves earlier values - measured at up to 0.19% over 1,500 SPY observations on `arch==8.0.0`, and at up to 0.9% of a security's largest value over 38 securities in sp500_equity_option_analytics' own section E. Every emitted value here inherits that through `variance`, so the refit schedule closes the parameter channel and leaves the conditioning channel open.

`garch11_conditional_volatility` on #764 cannot replace the call as written: it is symmetric GARCH(1,1) with no gamma, and this case study fits GJR (`o=1`). Three of the four GARCH case studies do.

The fix is one shared gamma-aware recursion rather than a fourth private one. There are already three hand-rolled causal GJR filters in the tree - `sp500_equity_option_analytics/04:388 causal_gjr_filter`, `sp500_options/04:506 causal_gjr_garch_filter`, and the half-hand-rolled step here. Sequence: #764 lands the helper, a small follow-up adds `gamma` to it with `gamma=0.0` asserted bit-identical for etfs, then this notebook calls it and the two private filters are deleted as their case studies come up.

Found by agents-ce, who read the four `arch_model` specs rather than the descriptions of them.

stefan-jansen:
Undrafted. The emission path no longer reads `arch`'s result object, which is the defect this PR was drafted over.

**One claim in the shared helper's docstring was wrong, and correcting it changes what this PR asserts about crypto.** `ARCHModel.fix` seeds from `self.resids(self.starting_values())` - residuals at the *estimated* mean, not at the parameters being fixed. crypto fits `mean="Zero"`, so nothing is estimated, the residuals are the returns, and the seed does **not** move when the sample is extended. On a healthy fit the old and new paths are bit-identical.

What remains under a zero mean is `variance_bounds`, whose per-row EWMA is clamped by two whole-sample quantities - `np.var(resids)/1e8` below, `1e7 * (1 + max(resids**2))` above. They sit six orders of magnitude apart, so they reach an emitted value only when the variance reaches a clamp, which is what a degenerate fit does. Measured with `omega=1e-8, alpha=0.02, gamma=-0.30, beta=0.90` and a cascade 500 bars past the cut: more than 400 shared emitted rows move, by up to 64%.

So the channel opens on exactly the fits that were also being silently clipped. That is the worse of the two failure modes, because spot-checking a well-behaved series cannot find it.

**A walk-level truncation test cannot probe this**, and that was measured rather than assumed. `walk_forward_feature` calls `apply(model, X[:emit_end])` with `emit_end` from `refit_boundaries`, which depends only on `burnin` and `refit_every` - so a completed block receives a byte-identical array in a truncated run and a full one. An earlier draft of the test file did truncate the walk and reported 420 shared rows with 0 differences against the very path this PR removes. The five tests are written against the emission path directly.

Depends on #770 for the gamma-aware helper; that branch is merged in here.

PR #769, opened 2026-09-05
CI: retry the container git install instead of failing on one apt hiccup

stefan-jansen:
Every job that runs inside the `ml4t` image installs git before checkout, and until now a single `apt-get update` failure against a Debian mirror ended the job at a step that has nothing to do with the change under test.

On 2026-09-05 that took out five case-study jobs across three unrelated pull requests inside one four-minute window:

| PR | job | failed step |
|---|---|---|
| #764 | `cs-fx_pairs` | `Install git` |
| #767 | `cs-cme_futures`, `cs-crypto_perps_funding`, `cs-fx_pairs` | `Install git` |
| #768 | `cs-us_firm_characteristics` | `Install git` |

All eight of those jobs are green on `main` at 1f245ced, and `Install git` was the only failed step in each. Each one costs a full notebook re-run to recover.

Eleven steps across `test.yml`, `full-test.yml` and `weekly-external.yml` now retry three times with a 15/30/45s backoff and fail loudly with an `::error::` annotation if all three fail, so a real breakage still stops the job and names itself.

`test-benchmark`'s copy ended in `|| true`, which swallowed the failure and let the job continue to a checkout that then failed for an unrelated-looking reason. The retry replaces it; the swallow is gone.

### DIFF
--- a/.github/workflows/full-test.yml
+++ b/.github/workflows/full-test.yml
@@ -41,7 +41,15 @@ jobs:
 
     steps:
       - name: Install git
-        run: apt-get update -qq && apt-get install -y -qq git openssh-client >/dev/null
+        run: |
+          # A transient apt mirror failure used to kill the whole job.
+          for attempt in 1 2 3; do
+            apt-get update -qq && apt-get install -y -qq git openssh-client >/dev/null && exit 0
+            echo "apt-get failed (attempt $attempt of 3); retrying in $((attempt * 15))s"
+            sleep $((attempt * 15))
+          done
+          echo "::error::apt-get could not install git after three attempts"
+          exit 1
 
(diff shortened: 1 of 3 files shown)


stefan-jansen:
`main` is red from this too, not just the three pull requests.

Run [33936111700](https://github.com/stefan-jansen/machine-learning-for-trading/actions/runs/33936111700) at `1f245ced` failed `ch11-12`, `ch24` and `ch14-15`, and `Install git (for checkout inside container)` is the only failed step in each. That brings the window's total to eight jobs across four runs, all one cause and none of them a defect in the code under test.

Reruns of the five case-study jobs are in flight and are expected to pass without a code change, which is the evidence that the step, not the change, is what failed.

PR #771, opened 2026-09-05
nasdaq100: label the interval the strategy holds, and purge the bar the VWAP actually reads

stefan-jansen:
Eleven commits that have been sitting on `cs6/nasdaq100-vwap-labels` with no pull request. nasdaq is the next case study to run end to end, and this is the label definition it would run against.

## The horizon and the purge were the same number, and they are not the same thing

`resolve_label_horizon` falls back to the **buffer** when no horizon is declared. So the buffer was doing two jobs at once: stating how far ahead the label looks, and stating how much of the training window to purge. Widening one silently lengthened the other, and `fwd_ret_15m` would have become a sixteen-minute return that every consumer still called fifteen.

`labels.horizons` is now declared explicitly, per label.

## The purge is the horizon plus one bar, and that is not conservatism

The entry leg is the VWAP of the bar **after** the decision and the exit leg is H past the entry, so a label at $t$ consumes a quote at $t + H + 1$ - one bar past the horizon its name states. A buffer of H leaves the last bar of every training window inside the first validation label's window.

| label | horizon | buffer |
|---|---|---|
| `fwd_ret_15m` | 15min | 16min |
| `fwd_ret_5m` | 5min | 6min |
| `fwd_ret_60m` | 60min | 61min |
| `fwd_dir_15m` | 15min | 16min |

## Also here

- Each unlabelled row is asked **why** it is unlabelled, because a VWAP is not a midpoint and a missing one has more than one cause.
- The horizon split is checked per declaration, so the guard fails on the revert it names rather than passing vacuously.
- `03_financial_features` is reduced like every stage around it.
- `tests/test_nasdaq100_label_execution_basis.py`, 321 lines, 13 tests - all passing.

## The merge

`main` moved 27 commits underneath this branch. Only the generated `.ipynb` files conflicted; every `.py` merged cleanly, so the notebooks were regenerated from the merged source and their provenance cleared. None of them had been executed since the changes were written, so nothing claiming a result was discarded - `git log` shows the register carries these as `fixed`, not `done`, and the nasdaq registry holds zero backtest rows and zero causal rows.


stefan-jansen:
Closing: every change on this branch is already on `main`.

`config/setup.yaml` on `main` carries `labels.horizons` per label and the H+1 buffer (16min / 6min / 61min), and `tests/test_nasdaq100_label_execution_basis.py` is present. Those landed as #737 and #756.

What the branch would still do is revert. Its merge of `origin/main` at 2c6cb990 resolved against `main` on four files, so merging it would:

- strip the executed outputs from `04_model_based_features.ipynb` (51 -> 0) and `05_evaluation.ipynb` (26 -> 0), which came from the stage 01-05 re-run in #756;
- delete `tests/test_sync_notebooks_prose_drift.py` and `tests/test_runtime_without_posix_resource.py`;
- roll back `reader-install.yml`, `case_studies/utils/runtime.py`, `scripts/sync_notebooks.py` and `uv.lock`.

Net against `main`: 778 insertions, 20,217 deletions. Nothing here is worth recovering.

PR #787, opened 2026-09-05
Fold 0 is the earliest window: adopt ml4t-diagnostic 0.1.4

stefan-jansen:
`ml4t-diagnostic` 0.1.4 constructs the backward validation windows from the held-out test boundary and then emits them chronologically, so a fold id now increases with time. The window set is unchanged - the library's test pins the `(train_start, train_end, val_start, val_end)` tuples across the change - and only the order and the ids attached to it moved. This adopts that release.

## Why the adoption is safe to attempt

`_assert_newest_first` existed for exactly this. With 0.1.4 installed and the guard untouched, `generate_cv_splits` raises on etfs with `val_starts` running 2015-12-30 to 2022-12-30, which is the failure it was written to produce rather than let forty call sites quietly mean the opposite. That is also what turned `main` red on `test-unit` before #784: the hand-built venvs resolved `>=0.1.2` to whatever had just published.

The guard is now `_assert_chronological` and asserts oldest-first. The lock and the four declared pins move to `>=0.1.4`; the workflow's install lines are deliberately untouched, because after #784 they take their versions from `uv.lock`.

## The sweep

57 positional fold sites in 23 files, each read for which fold it means rather than remapped. Three were order-dependent and silent:

- `15_causal_estimation/05_momentum_causal_trading.py` took `splits[0]["val_start"]` as the holdout boundary and `splits[0]` as the test period, under a comment stating that `splits` is reverse-chronological.
- `20_strategy_synthesis/holdout.py` measured the retrain window length from `splits[0]`, and fold windows differ in length, so it would have matched the shortest rather than the one ending nearest the holdout.
- `us_equities_panel/04` built the holdout fold's `train_start` from `folds[-1]["train_start"]` - the earliest CV fold before, the latest after, so the holdout would have fitted its transforms over a shortened history while claiming full history. Found by the pre-push review, not by the grep.

Plus one that only misreports: `fx_pairs/04` labels `folds[-1]` the oldest window when printing Kalman burn-in loss.

The remaining 53 are position-in-a-one-element-list (`gbm.py`, `linear.py` build `[split]` and then index it) or a demonstration picking any fold. A teaching notebook showing "a fold" shows a different one after this change, which is not a defect.

## What is deliberately not here

- `us_firm_characteristics/config/cv_config.json` stores its splits newest first and is now refused by the guard. Renumbering it is an identity change for the 164 runs registered under it and lands separately, with the `04_evaluation.py` print that has the same shape.
- Nine `01_feasibility_analysis` notebooks describe the old numbering in prose and label their fold figures with ids that now count the other way. Prose and figure have to move together, so they land with their re-render rather than as a prose edit against a stale figure.

## Verification

282 passed, 26 skipped across `test_cv_splits`, `test_holdout_boundary`, `test_cv_window_fold_splits`, `test_variant_fold_geometry`, `test_folds`, `test_holdout_cv_derivation`, `test_fold_resolution`, `test_research_models` and `test_ci_venvs_follow_the_lock`, against 0.1.4 installed from the bumped lock. Three order assertions in those tests moved with the change; one, `crypto_purges_variant_endpoint_at_holdout`, was reading `splits[0]` where it meant the fold at the holdout boundary and now calls `most_recent_split`.

Three notebooks have their outputs cleared rather than kept, because their executable source changed and none has been re-executed.


stefan-jansen:
Superseded by #791, which carries these same three commits plus the us_firm_characteristics fold renumber on top. Branched from this one before I rebased it, so the SHAs differ but the content does not: 07563680/d544713c/8f342da1 here are d2aa3d2b/ca9aabb8/23f2c603 there. Closing to leave one fold-numbering PR rather than two. The branch stays; nothing is deleted.

PR #804, opened 2026-09-06
fx_pairs: what 12_model_analysis' three diagnostics are for, and what actually blocks 19_strategy_analysis

stefan-jansen:
Two prose commits that were finished and unpushed when the fx session ended. Rebased off the merged #801 so this carries only its own two commits.

- `12_model_analysis`: says why fold spread, family agreement and conformal width matter, rather than naming them and moving on.
- `19_strategy_analysis`: says what actually blocks the notebook under a preview population, replacing an explanation that named the wrong cell.

Prose only. No executable source moved, so nothing re-runs for this.


stefan-jansen:
Duplicate of #805, which points at the same commit 8a53d24f - agents-95 and I recovered the same two unpushed fx commits onto the same base concurrently and got identical hashes. Closing this one; #805 is the live PR. Branch `cs/fx-pairs-notebooks` deleted.

PR #815, opened 2026-09-06
nasdaq100_microstructure: the nine stage-06+ notebooks that had no smoke declaration

stefan-jansen:
nasdaq was the worst-covered case study under criteria 11-13, at 6 of 15. This declares the other nine: **one measured configuration, two exemptions, six blocks.** Every entry records something measured on 2026-09-06 rather than something the notebook looks like it should do.

## The one that runs

`12_causal_dml`, `EXECUTION_TIER=preview` with `MAX_SYMBOLS=12, CV_FOLDS=2, MAX_SAMPLES=20000, N_PLACEBO=12`: **39 s, 8.7 GB `Maximum resident`, exit 0.** The `causal_runs` row it writes carries a computed `refutation_p` over 12 successful placebo draws - above `MIN_PLACEBO_DRAWS`, not merely at a number that parses - so the table the full run writes to has a real row in it.

`max_samples` is the lever. 20,000 rows of a one-minute panel is about four sessions across twelve names, and the fit is 3.2 s of the 39; the rest is reading the panel.

## The two that are exempt

`18_holdout_predictions` and `19_holdout_backtest`, for the reason every case study's holdout pair is exempt: `case_studies/research/models.py:75` refuses a training spec whose `execution_tier` is not canonical, and 19 reads the prediction set 18 would have produced.

## The six that are blocked

One reason with two halves, both measured.

**They are not on the research boundary.** 14, 15, 16, 17 and 20 call no `open_study` at all - they import `run_backtest` and open `run_log/registry.db` with `sqlite3` directly - and `13_model_analysis` reads through `Study.at`, which takes one root and performs no activation. None of the six declares `EXECUTION_TIER` or `WORKSPACE`, so there is no preview tier to ask for, and a reduced run would register into the case study's own artifacts. That is the production path, which is what the tier boundary exists to keep a smoke run out of.

**There would be nothing for them to read.** In the registry these notebooks select from: 13 `training_runs`, and 0 rows in `prediction_sets`, `backtest_runs`, `decision_artifacts`, `candidate_sets` and `fold_metrics`. The one official population, `nasdaq100_microstructure-linear-validation-v1`, declares 61 prediction members and none of them exists.

### The first half is a gap in the standard, not six missing declarations

Do not try to write these six declarations. A notebook with no research boundary has no preview tier to declare, so the smoke standard cannot cover it by construction: there is no parameter to set that would keep a reduced run out of the case study's own artifacts. `tests/test_smoke_configs.py::test_every_declared_parameter_reaches_its_notebook` enforces exactly this - an `EXECUTION_TIER: preview` written against `14_backtest` is a parameter that reaches nothing, and the test refuses it.

So the sequence is: put the notebook on the research boundary first, then declare its smoke configuration. Six of nasdaq's are the instance that surfaced it; the standard's own coverage question is the general form.

Each entry names its own completion issue (466, 467, 468, 469, 470, 471) rather than a shared one, because finishing that notebook is what unblocks it.

## Verification

`tests/test_smoke_configs.py`: 74 passed. The nine lines come out of `UNDECLARED`, which is a ratchet - the test asserts the missing set is exactly that list, so nothing can rejoin it silently.

https://claude.ai/code/session_01P3GtYYaqpHTe5itLUYXomS


stefan-jansen:
A note on this PR's green tick, for anyone reading it later as evidence.

The merge state was CLEAN, but only **12 checks ran**. This PR touches `tests/smoke.yaml` and `tests/test_smoke_configs.py` and nothing else, so path filtering skipped every chapter and case-study job - `ch11-12`, `ch14-15`, `ch26` and all nine `cs-*` jobs among them. None of them ran here.

At merge time `main` was red on `ch11-12`, `ch14-15` and `ch26`. This PR's CLEAN is not evidence that any of them passed, and it does not contradict that state. What it does say is that the smoke declarations parse, that every declared parameter reaches its notebook, and that the `UNDECLARED` ratchet is exactly the set it should be - which is the whole of what `test_smoke_configs.py` is for.

The three reds are tracked elsewhere: `ch11-12` and `ch14-15` clear with the CI fixture regeneration; `ch26` does not and is a separate cause (`01_drift_monitoring` on `DataFrame is empty`, `05b_feast_live` on a two-source join that finds no overlapping event, both at cell 17, neither naming a fold).

Issue #300, opened 2023-08-03
problem in Long-Short Strategy, Part 1: Preparing Alpha Factors and Features


ishikabansal77:
def compute_bb(close):
    high, mid, low = BBANDS(close, timeperiod=20)
    return pd.DataFrame({'bb_high': high, 'bb_low': low}, index=close.index)

def compute_bb(close):
    high, mid, low = BBANDS(close, timeperiod=20)
    return pd.DataFrame({'bb_high': high, 'bb_low': low}, index=close.index)
​
bb_data = prices.groupby(level='symbol')['close'].apply(compute_bb)
prices = pd.concat([prices, bb_data], axis=1)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[33], line 1
----> 1 prices = pd.concat([prices, bb_data], axis=1)

File ~\anaconda3\envs\baclass\lib\site-packages\pandas\core\reshape\concat.py:372, in concat(objs, axis, join, ignore_index, keys, levels, names, verify_integrity, sort, copy)
    369 elif copy and using_copy_on_write():
    370     copy = False
--> 372 op = _Concatenator(
    373     objs,
    374     axis=axis,
    375     ignore_index=ignore_index,
    376     join=join,
    377     keys=keys,
    378     levels=levels,
    379     names=names,
    380     verify_integrity=verify_integrity,
    381     copy=copy,
    382     sort=sort,
    383 )
    385 return op.get_result()

File ~\anaconda3\envs\baclass\lib\site-packages\pandas\core\reshape\concat.py:563, in _Concatenator.__init__(self, objs, axis, join, keys, levels, names, ignore_index, verify_integrity, copy, sort)
    560 self.verify_integrity = verify_integrity
    561 self.copy = copy
--> 563 self.new_axes = self._get_new_axes()

File ~\anaconda3\envs\baclass\lib\site-packages\pandas\core\reshape\concat.py:633, in _Concatenator._get_new_axes(self)
    631 def _get_new_axes(self) -> list[Index]:
    632     ndim = self._get_result_dim()
--> 633     return [
    634         self._get_concat_axis if i == self.bm_axis else self._get_comb_axis(i)
    635         for i in range(ndim)
    636     ]

File ~\anaconda3\envs\baclass\lib\site-packages\pandas\core\reshape\concat.py:634, in <listcomp>(.0)
    631 def _get_new_axes(self) -> list[Index]:
[...]

PR #816, opened 2026-09-06
us_equities_panel: the fifteen stage-06+ notebooks read, not swept

stefan-jansen:
Fifteen of `us_equities_panel`'s seventeen stage-06-and-later notebooks had been swept for banned vocabulary, not read. This is the reading, and the fixes it produced. Nothing was executed.

`20_strategy_analysis` is deliberately not read here: its reading needs `case_studies/utils/paired_metrics.py` read alongside it, and ml4t/agent-workspace#1049 is still open on the inference that file states. One mechanical fix reaches it - its figure had no alternative text - and the rest of it stays owed.

## The blocker this reading found

**`16_backtest` cannot freeze its baseline sets, and everything downstream of it reads names it was meant to create.** Its `PREDICTION_SET_NAMES` declares sixteen sets, six of them latent, and section 6 freezes one set per label over all of them with no comparison contract. `CandidateSet.create` then requires every protocol field to match, `Result.protocol()` includes `feature_artifacts`, and the latent adapter builds that field from a different object than the other nine producers. Every label's pool holds both kinds, so every freeze raises.

That is ml4t/agent-workspace#891, open, `lane:infra`. Its body named `20_strategy_analysis` as the first mixed pool; `16_backtest` is earlier in the chain and hits it first, which I have added to the issue. **The notebook is not changed**: #891 says uep was told to report the refusal rather than declare `feature_artifacts` comparable, and the refusal is the evidence it is waiting for. So stages 16 through 20 are blocked on #891 for execution regardless of anything in this PR.

## What repeated across notebooks

- **Nine notebooks carried an eight-line comment explaining `open_study` by describing what `Study.open(workspace=...)` does wrong**, citing `workspace.py:274` and `_ensure_input_link`. A note to whoever maintains the workspace code, in a textbook, against T10. `08_tabular_dl` already had the reader-facing version of the same paragraph; it is now in all ten. Two more of the same shape went with it: an eleven-line note at `16` and `18`'s freeze sites, and `17` and `18`'s price-cache note with five file:line references. No `.py:NNN` reference remains in any uep notebook.
- **"The downstream notebooks do not open populations" is false**, in `06`, `07` and `08`. `15_model_analysis.py:211` opens `OfficialPopulation.one` for every name in `OFFICIAL_POPULATION_NAMES`, which holds the checkpoint population each model notebook creates. True of `16_backtest`, false of `15`.
- **`06` and `07` said the downstream notebooks would name "four sets that nothing produces."** Both freeze two sets on each of three labels. Six.
- **`09`, `10` and `11` recounted a sibling's fixed defect** - "`11_dl_tsmixer` carried a pinned symbol whitelist to avoid it, which repaired one panel" - and wrote `min_obs` into prose where `08` says "a minimum number of names quoted on a date."
- **Four matplotlib figures called `fig.show()`**, in `15`, `17`, `19` and `20`, which emits an `` with no alternative text. All four now use `show_with_alt`. That also closes uep's share of #300 at the mechanism: `fig.show()` under `MPLBACKEND=Agg` is what suppressed the PNG, and `display()` does not. Commented on #300.
- **An empty code cell in each of `08`, `09`, `10`, `11`**, and `Study`, `REPO_ROOT`, `warmup_periods_for`, `load_configs` imported and never used across eleven notebooks.

## What was specific to one notebook

- **`12_dl_weekly` reported a maximum as a result.** `run_dl_cv`'s `best_ic` is `max()` over each configuration's checkpoint ICs, and the summary table is three of those compared against a registry query. `07` through `11` each spend a paragraph forbidding exactly that reading. **No number moves** - Table 13.5 depends on them - and the section now says what they are. Also in `12`: the NLinear description that was corrected in `09` and survived here; a `/tmp` log written five times and read by nothing (C13); a hand-written SQL join over three registry tables, replaced by the registry's accessors (C12); blanket warning suppression; and mid-notebook imports.
- **Three notebooks argued from a figure they never drew.** `10_dl_lstm` names the learning curve as its evidence, `14_causal_dml` describes the refutation as a distribution and shows only its p-value while the hundred draws sit in `metrics["placebo_effects"]`, and `18_risk_management`'s whole case is how the effect moves across fourteen thresholds. All three now draw it.
- **`15_model_analysis`'s IC chart was `0.28 * height` inches tall with one labelled row per candidate**, over a population of roughly 588. Replaced by two strip panels - point estimates and interval half-widths, one column per label and family - which is what the section's two questions need and what scales.
- **`13a_pca` and `13b_ipca` declared the factor count three times and bound none of them.** The preset, a notebook parameter, and a literal in the publication gate; the request overrode the preset unconditionally, so editing `config/pca/pca.yaml` changed nothing while the gate kept publishing. The preset now decides, and `OVERRIDES` still wins and still blocks publication.
- **`15`'s middle and `19`'s takeaways were written for a reviewer** rather than a reader, and are rewritten from material already in their own preambles.
- **A garbled sentence I introduced in `18`'s `overlay_effect` comment** on the merged branch ("A row identical on both changed no / neither of the two statistics") is fixed.

## Withdrawn on reading the code

I raised `Study.open` on `13_latent_factors`' and `15_model_analysis`' canonical path as inconsistent with the sibling comment saying never to call it directly. `Study.open(case_study)` with no workspace is the read-only open; `open_study` on that path calls `Study.regenerate`, which opens read-write and requires symlinked artifact directories. Both notebooks write nothing, so both were right. The over-broad comment is gone with the T10 sweep.

## Verification

Static only, which is what this diff is. `ruff` clean; 301 tests pass across `test_results_cell_cap`, `test_us_equities_panel_named_sets`, `test_us_equities_panel_model_based_schedule`, `test_smoke_configs` and `test_latent_factors_no_leak`; results-cell counts still match the N7 ratchet exactly; every executed output in `13a`, `13b`, `18` and `19` survived the pair sync.

Prose depth, `agents/scripts/prose-depth.py` against this worktree: **16 of 17 over the 92-line floor, from 3 of 17**. `13_latent_factors` falls 73 to 70 because two padding sentences describing its own cell came out, which is the right direction under "cut for padding, never for length"; it stays the ruled exception.

## Still owed after this

- `20_strategy_analysis`' section B reading, behind #1049.
- Execution of stages 16 through 20, behind #891.
- Smoke configurations: every uep stage-06+ notebook is on `test_smoke_configs.py`'s UNDECLARED ratchet, and declaring one needs a measured run.
- The N7 results-cell debt in `16`, `17`, `18` and `19`, which the ratchet says is paid off when each is next re-run.
- The per-control trigger count `18_risk_management` wants, ml4t/agent-workspace#1051.

https://claude.ai/code/session_014SP3KUnd2D1Hjc4p8beVzN


stefan-jansen:
**Three corrections RoboRev caught in this PR's own new prose, recorded because the pattern is worth naming.**

The gate ran four times. It passed every time with no blocking findings, and raised four Medium ones. All four were mine, all in sentences written during this reading, and all the same failure the reading was for: a claim asserting more than the computation supports.

1. `14_causal_dml` - I wrote that `naive_effect` is "the treatment regressed on the outcome ... what a reader would get by correlating momentum with the forward return". `causal.py:954-956` regresses the return on momentum and takes the slope, which is not the correlation. Both halves wrong.
2. `12_dl_weekly` - I wrote that both weekly arms carrying a maximum over ten checkpoints leaves their comparison "not distorted by it". The same operation is not the same bias; it depends on how the ten vary and how correlated they are.
3. `18_risk_management` - I wrote that a curve improving toward the loosest threshold says the overlay is worthless, with the unprotected book nowhere on the page, and that a flat curve means the control never binds - which is the inference the `overlay_effect` cell four sections above refuses to draw. The figure now carries the unprotected reference and the prose says "unchanged measured Sharpe".
4. `12_dl_weekly` again - fixing (2), I replaced it with "a gap smaller than the spread of the curve behind them is not a difference between formulations", treating a learning curve's range as an error bar. It is systematic, not sampling variation.

The shape of my own error is consistent and worth stating: **on removing a false claim I reached for a replacement heuristic instead of stating the limit.** Twice in the same paragraph. The version that finally passed says the table cannot settle the comparison and names what would - a paired comparison on shared dates with an interval - rather than offering a rule of thumb in its place.

That is an argument for the gate rather than against the reading. Every one of these would have shipped as a confident sentence in a textbook.

PR #823, opened 2026-09-06
Name the notebook on the four etfs model requests the chain has not reached

stefan-jansen:
A registered training run should say which notebook produced it. `tests/test_model_registry.py::test_model_notebook` checks exactly that and is in `.github/ci/unit-test-quarantine.txt`, where the entry reads "nothing runs it".

The mechanism already exists end to end: `model_requests(..., notebook=)` reaches every family runner, and each writes `runtime_provenance["notebook_path"]` - `linear.py:189`, `gbm.py:1524`, `tabular_dl.py:141`, `deep_learning.py`, `latent_factors/adapter.py:162`. Only `us_firm_characteristics` passes it, so 54 of the 61 registering notebooks write rows that name nothing.

## Why only four notebooks

`notebook_path` is in `_V2_PROVENANCE_FIELDS`, so no training identity moves and nothing refits. But the argument goes in a code cell, and the provenance gate keys on the paired `.py` (`source_py_blob`), so threading it into an already-executed notebook makes that notebook's stamp stale and costs it a re-run.

Running the gate over all 54 gives 39 stamped and 15 unverified. These four carry no stamp and sit at jobs 7 to 10 of `prod-etfs-2`, which is on job 3, so they are threaded ahead of their own execution and cost nothing.

The other 50 are not free in the same way:

- **9 cannot be threaded at all.** `Study.causal` passes `**request` into `CausalRequest`, which has no notebook field, so every `*_causal_dml` notebook needs a library change before its call site can name anything.
- **10 belong to two active case studies** (`nasdaq100_microstructure`, `us_equities_panel`) and land with work those sessions are already doing.
- **The rest are stamped**, and are threaded as their chains reach them rather than by re-running a notebook to add a provenance field.

The check stays quarantined until a regenerated case study's `cs-<id>` passes it. Un-quarantining now would put a red check on `main` on landing day.

Verified: `notebook_provenance.py check` reports the four as unverified with 0 stale, and `uv run ruff check` passes.

https://claude.ai/code/session_016Smmj8LBZc1vzgXHQDZsmX

### DIFF
--- a/case_studies/etfs/10a_dl_nlinear.ipynb
+++ b/case_studies/etfs/10a_dl_nlinear.ipynb
@@ -345,6 +345,7 @@
     "    execution_tier=EXECUTION_TIER,\n",
     "    overrides={\"device\": device},\n",
     "    preview_reductions=PREVIEW_REDUCTIONS,\n",
+    "    notebook=\"10a_dl_nlinear\",\n",
     ")\n",
     "resolved = tuple(request.resolve() for request in requests)\n",
     "\n",
--- a/case_studies/etfs/10a_dl_nlinear.py
+++ b/case_studies/etfs/10a_dl_nlinear.py
@@ -241,6 +241,7 @@ requests = model_requests(
     execution_tier=EXECUTION_TIER,
     overrides={"device": device},
     preview_reductions=PREVIEW_REDUCTIONS,
+    notebook="10a_dl_nlinear",
 )
 resolved = tuple(request.resolve() for request in requests)
 
--- a/case_studies/etfs/11a_pca.ipynb
+++ b/case_studies/etfs/11a_pca.ipynb
@@ -324,6 +324,7 @@
     "    execution_tier=EXECUTION_TIER,\n",
     "    overrides=overrides,\n",
     "    preview_reductions=PREVIEW_REDUCTIONS,\n",
+    "    notebook=\"11a_pca\",\n",
     ")\n",
     "resolved = tuple(request.resolve() for request in requests)\n",
     "\n",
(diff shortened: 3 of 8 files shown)


stefan-jansen:
## Correction: `10a_dl_nlinear` is inert, the other three record

The description says these four make their rows name themselves. Three do. `10a_dl_nlinear` does not, and the reason is a library gap I had not found when I opened this.

Four family modules read the boundary's notebook key:

```
case_studies/utils/gbm.py:1820                    request.get("notebook")
case_studies/utils/linear.py:571                  request.get("notebook")
case_studies/utils/tabular_dl.py:441              request.get("notebook")
case_studies/utils/latent_factors/adapter.py:515  request.get("notebook")
```

`case_studies/utils/deep_learning.py` does not, and has zero occurrences of `notebook_path`. Its `resolve_model_request` (`:413`) is what the shared boundary calls for the sequence families, and it never reads the key. So `study.model(..., notebook=)` accepts the field, `as_dict()` carries it, and the sequence resolver drops it.

The module does take a `notebook` argument at `:1892`, forwarded as `entry_point=notebook` at `:1915` - but that is the older column mechanism on a different entry point, not the one the boundary uses.

So this PR's effect is:

| notebook | family | records |
|---|---|---|
| `11a_pca` | latent_factors | yes |
| `11b_ipca` | latent_factors | yes |
| `11c_conditional_autoencoder` | latent_factors | yes |
| `10a_dl_nlinear` | deep_learning | **no** - accepted and ignored |

Leaving the fourth in rather than reverting it: it is one keyword argument, it is correct at the call site, and it starts recording the moment the DL resolver reads the key. Removing it would only mean threading it again later.

The library fix belongs on ml4t/agent-workspace#1059 with the causal one, not here - five chains are executing against `deep_learning.py` right now.

PR #830, opened 2026-09-06
cme_futures: a fold-free stage-04 artifact, and the eighteen stage-05+ notebooks executed against it

stefan-jansen:
`04_model_based_features` regenerates `model_based.parquet` without the `fold`
column, so a downstream scan selects rows by timestamp instead of inheriting a
fold assignment. That invalidated every result fitted against the old artifact,
so the run log was reset (a baseline was recorded first) and stages 05 through
19 were re-executed in order on the current sources.

**All eighteen exited 0.**

Registry after the run: 101 training runs across five families (linear 56,
gbm 30, tabular_dl 6, latent_factors 5, deep_learning 4) and 1,151 backtest
rows - 992 baseline, 120 allocation, 28 risk overlay, 11 cost sensitivity,
1 holdout.

Both declared labels have backtest rows, so `scripts/label-coverage.sh` passes
without a second pass:

| label | | backtests | prediction sets |
|---|---|---|---|
| `fwd_ret_5d` | primary | 570 | 248 |
| `fwd_ret_21d` | variant | 582 | 249 |

`10b_stochastic_discount_factor` needed four attempts. The first three died on
the latent-factors aggregator reading an empty catalog, because
`10_latent_factors` was queued ahead of its component notebooks rather than
after them. That is a chain-ordering defect, not a notebook defect, and it is
fixed in the chain.


christy108:
PLEASE UNSUBSCRIBE ME FROM THESE FUCKIKG EMAILS

PR #832, opened 2026-09-07
ch26: read uep's model-based features by date, not by a fold column it no longer has

stefan-jansen:
`us_equities_panel` stage 04 writes `model_based.parquet` **fold-free by design** - a stock-date carries one value whichever fold later reads it, stated in its own prose at `04_model_based_features.py:466-469`. Three `26_mlops_governance` notebooks still keyed on a `fold` column that artifact does not have.

| notebook | was |
|---|---|
| `01_drift_monitoring` | `pl.max("fold")`, `pl.col("fold")` |
| `05_feast_feature_store` | `pl.max("fold")`, `pl.col("fold")` |
| `05b_feast_live` | `pl.max("fold")`, `pl.col("fold")` |

## Why this is urgent rather than merely open

`05_feast_feature_store` **passes in CI today**, and it cannot pass against a fold-free frame: `pl.max("fold")` raises `ColumnNotFound` before any window logic runs. That it passes is evidence CI still installs an older fixture that carries the column. Every `model_based.parquet` under `~/ml4t/test-data/intermediates/` is already fold-free, so the moment that fixture publishes, all three fail on a missing column and `05` regresses from green. The regeneration does not clear ch26; it makes it worse.

## The change

All three now filter the artifact **once**, with `pl.any_horizontal` over the union of the walk-forward validation windows plus the sealed holdout, instead of concatenating one fold-filtered frame per window.

The concat is why the duplicate-key assertion could not mean much: overlapping windows would double-count, and the assertion would report it only after the fact. Under a single filter a duplicate cannot arise, so the assertion goes back to stating the thing it is for - that the artifact is unique on `(symbol, timestamp)`. The windows are disjoint today, so this was latent rather than live, which is the reason to remove the shape while the block is already being rewritten rather than to carry it forward.

## The pinned dates go too

Same defect, one level down. `01` pinned `REFERENCE_START = 2015-01-01`; `05` and `05b` pinned `TRAINING_START = 2015-10-01`, `TRAINING_END = 2015-12-30`, `AS_OF_DATE = 2016-01-04`. None was an independent choice - `TRAINING_END` equalled the then-final validation fold's `val_end` to the day - and #819 restored uep to 16 folds, which shifts every boundary. A window written as a literal is a claim about the fixture's geometry that nothing checks, and it silently becomes false.

Each is now derived from `generate_cv_splits` and stays overridable by passing a date. Both `sample_assets` functions ranked liquidity over a pinned December and now rank over the tail of the derived window.

The latest window is chosen with `max()` over the spans rather than by list position: `ml4t-diagnostic` 0.1.4 reversed fold numbering, so "the last element" is exactly the assumption that just moved.

`05`'s fail-closed guard (`TRAINING_END` must precede the sealed holdout) now runs **after** the derivation, so it still covers a hand-pinned override, which is the case it exists for, with the symmetric `TRAINING_START <= TRAINING_END` beside it.

## Verification

    3 passed against the CI fixture (tests/test_chapter_notebooks.py)
    ruff check / ruff format: clean
    provenance gate: 0 stale
    grep -c '"fold"' across the three: 0

One trap worth recording for anyone reproducing this: my worktree venv held `ml4t-diagnostic` 0.1.2 while `uv.lock` pins 0.1.4, and the first run failed with a `RuntimeError` about folds not ordered oldest first that reads exactly like a data defect. It was the venv. `uv sync --frozen --extra dev --extra live` fixes it; plain `--frozen` removes pytest.

## The three land cleared, deliberately

Per `AGENTS.md` "Environment": the code is correct but not yet executed, so the notebooks are committed with `notebook_provenance.py clear` - no outputs, no stamp, claiming nothing. This is a recorded state, not a failed run.

They are not re-executed here because they **cannot** be, against production: `~/ml4t/artifacts/case_studies/us_equities_panel/features/model_based.parquet` is from 2026-08-10 and is still fold-replicated (December 2015: 99,421 rows against 50,869 distinct keys, **1.95x**), so the new duplicate-key assertion correctly fires against it. That is the code working - the production artifact is stale relative to stage 04 on `main`. Executing against the CI fixture instead would replace production-scale reader figures with trimmed ones under notes that say "production", which is worse than no figures.

Re-execute all three against production and stamp them as production runs once uep stage 04 reaches the production store. The measurement is recorded on ml4t/agent-workspace#1020, which this corroborates on a second case study (nasdaq 2.0x, uep 1.95x).

Refs ml4t/agent-workspace#1063, ml4t/agent-workspace#1020

https://claude.ai/code/session_01TvQUuNRz572ohmUwscwmzp


stefan-jansen:
One thing to carry into the production re-execution, from `agents-55`'s corpus measurement (ml4t/agent-workspace#1069).

`ml4t-diagnostic` 0.1.4 fixed `WalkForwardCV` numbering its folds backwards in time, released 2026-09-05, mid-wave. Of the 242 committed notebooks that mention folds, 89 render the chronological numbering and 153 the reversed one. **All four `26_mlops_governance` notebooks are in the reversed 153** - the three here plus `06_mlflow_experiments`, executed 2026-08-10 and 2026-09-04.

So when these three are re-executed against production, their rendered fold ids will flip to chronological. That is the correct direction and not a regression, but it will read as an unexplained change to anyone diffing the render, so it belongs in the stamp notes at that point rather than being discovered later.

It is also an independent argument for landing them cleared: outputs written today would have been superseded on this axis regardless of the fold-column change this PR is about.

stefan-jansen:
**`ch26` was red on the first push, and the reason was worth the round trip: the assertion was right and the artifact CI installs is not.**

    Error: Cell 18 (AssertionError): model_based.parquet is not unique on (symbol, timestamp)

on all three. CI installs the published test-data fixture, which is still the **fold-replicated** one; only the local tree is fold-free. So the first version coupled this PR to the fixture publication: the code could go green only after the fixture landed, while the fixture was being held for the code.

`a738452b` breaks that. `collapse_fold_replication` is a no-op once the frame already has one row per key, and where it does not, it collapses **only after proving the collapse loses nothing** - `unique()` over every selected column has to reduce to exactly the distinct key count, which happens only when the replicated rows agree everywhere. A real per-fold difference raises rather than being reduced to whichever row sorted first. This is not "tolerate duplicates"; #1020 measured 0 of 22 columns differing between folds, and this checks that per read instead of trusting it.

Verified against three artifact shapes, built rather than reasoned about:

| fixture | result |
|---|---|
| fold-free (stage 04 today) | 3 passed |
| fold-replicated, values identical (what CI has) | 3 passed |
| fold-replicated, one feature genuinely differing | correctly refuses, naming the row and key counts |

The third is the one that makes the other two mean anything. **My first attempt at it passed, and passed for the wrong reason:** `ML4T_DATA_PATH` pointed at a symlinked `data/`, so `_resolve_intermediates_root` took `Path(data_path).parent` back to the real tree and the notebooks never read the doctored file. Both the "legacy passes" and "poisoned passes" results from that round were void. The harness had to be fixed before any of it counted.

stefan-jansen:
**Final state, and `ch26` is still red - deliberately, with a message that says what to do.**

The published CI fixture does not describe the geometry this case study declares:

    ValueError: model_based.parquet carries 17 folds and this notebook derived 16
    evaluation windows. They cannot be paired by date, so the artifact does not
    describe the geometry this case study now declares; regenerate it against the
    current stage 04.

All five **required** checks pass. `ch26` is not required.

## How this got to the right answer, since it took three rounds

**Round 1** asserted the artifact is unique on `(symbol, timestamp)`. True of stage 04's output, false of the published fixture, so all three went red on the assertion - correct behaviour against a stale artifact, but it coupled the PR to the fixture publication.

**Round 2** tried to collapse the replication, on the assumption that a fold-replicated artifact repeats the same value per fold. #1020 measured exactly that on nasdaq (0 of 22 columns differing). It is **false for this fixture**: of 22,890 rows over 11,467 keys, **22,576 disagree on a feature value**. Those folds are genuine vintages, not copies. The guard refused rather than averaging them away, which is the check earning its keep on the first real artifact it met.

**Round 3**, after RoboRev blocked the push on a HIGH finding worth quoting, because I would have shipped it:

> The legacy branch pairs stored fold IDs with newly generated fold IDs, although `ml4t-diagnostic` 0.1.4 reversed their chronological meaning.

Both ids exist, so the join succeeds and silently returns the wrong vintage for every date. My own legacy fixture could not have caught it - I built it with 0.1.4 ids, so it was self-consistent by construction, which is the same defect shape as testing a producer against a copy of itself.

`folds_by_coverage` now never reads what a fold id *means*. It orders the artifact's folds by the last date each covers, orders the derived windows by date, and pairs positionally - the one correspondence that holds under either convention. A fold count that disagrees with the window count is refused rather than zipped short, which is what fires above.

## Verification

| fixture | result |
|---|---|
| fold-free (stage 04 today) | 3 passed |
| fold-keyed, 0.1.4 forward ids | 3 passed |
| fold-keyed, 0.1.2 reversed ids | 3 passed |
| differential: same rows under both conventions | **True**, no duplicate keys |

The differential test pulls `folds_by_coverage` and `window_terms` out of the notebook with `ast` rather than reimplementing them, so it cannot pass by agreeing with a copy of the logic while the notebook drifts.

## What makes ch26 green

The regenerated fixture. Against a fold-free artifact the fold path is not taken at all, and these three read by date - which is the change #1063 asked for. Until then the red is accurate: the fixture on CI is two geometry changes behind what `setup.yaml` and `generate_cv_splits` now declare, and no notebook change can make reading it correct.

PR #852, opened 2026-09-08
Two chains read one universe each, and the model stages that bound them move with them

stefan-jansen:
Nine case studies audited for whether their label-taking stages read one universe. Two do
not, and in both the chain is not what binds.

## What was measured

| case study | chain caps | |
|---|---|---|
| sp500_equity_option_analytics | 21, 6, 5, 5 | three universes |
| us_firm_characteristics | 12, 12, 5, 5 | two universes |
| nasdaq100_microstructure | 12, 12, 12, – | consistent |
| us_equities_panel | 45, 45, 45, 45 | consistent |
| the other five | no `MAX_SYMBOLS` on the chain at all | |

seoa backtested 21 symbols and then priced and risk-checked 5 - not the panel the
backtests were run on. That is accretion rather than a considered set: #850 raised exactly
one of its four stages.

## The chain was not the binding constraint

`get_entry_schemes_for` filters against the width of the predictions a sweep **ranks**, so
the model stages bound it, not the notebooks that read the panel. Raising 11-14 alone would
have repaired the chain against a universe its inputs never produce, and the sweep would
still have registered nothing.

That is why `us_firm_characteristics`' model stages move with its chain. Counting the
fixture is what showed it - 400 symbols in the features, 200 in the folds - where reasoning
about the chain would have shipped a change that altered nothing and looked like it worked.

## us_firm at 41, and the shortfall it leaves

Long-short with `top_k_grid [5, 10, 20, 50]`, so `k` needs `2k` disjoint names. Estimated
from the stage times on CI run 34157373964, scaling with rows:

| width | concentrations | est. stage total | added |
|---|---|---|---|
| 5 | 0 of 4 | 144s | +0s |
| 12 | 1 of 4 | 184s | +40s |
| 20 | 2 of 4 | 229s | +86s |
| **41** | **3 of 4** | **470s** | **+327s** |
| 101 | 4 of 4 | 1158s | +1014s |

101 triples a job that runs 8m57s, and `08a_ipca` is 80% of the base before any scaling.
41 costs about +5.5 minutes and buys `k=20`.

**So `us_firm_characteristics` realizes three of the four concentrations it declares, and
`k=50` does not fit: two disjoint legs of fifty need 101 names and the model stages fit
41.** That sentence is in the `05_linear` entry and on ml4t/agent-workspace#1075, rather
than left for the next person to measure again.

`04_evaluation` and `09_causal_dml` stay at 5. Neither feeds the population `11_backtest`
ranks, and `05_linear` reads its configs from the shared catalog rather than from the
evaluation artifact - checked rather than assumed. A raise that touches stages the sweep
never sees is cost with no coverage.

## Two things this deliberately does not assume

**`08a_ipca` is iterative**, so twice the universe may cost more than twice the time, and
it dominates every row of that table. The run measures it against 114.9s at 20. If it
comes in materially above 2x, that changes what anyone should believe about the rest of
the ladder, not just that row.

**seoa's model stages inject `max_symbols: 5`** - measured through `injected_parameters`,
since `06_linear` and `07_gbm` declare no reduction and take the harness default. If its
sweep ranks predictions those stages produced, its cross-section is 5 and a long-only
`k=5` needs 6, so the chain raise realizes nothing on its own. Raising them is not the
answer 41 was: `11d_stochastic_discount_factor` measured **665s at 5 symbols**, so one
stage is 40% of a 26.5-minute job before anything scales, and four times the universe
makes it a job of its own. The caveat is in the entry rather than guessed at, and if the
run shows the sweep still empty that is the finding - it belongs with
ml4t/agent-workspace#1084 and #1087.

https://claude.ai/code/session_013HHT2sUgKgTC2GfpePdm4m

### DIFF
--- a/tests/overrides.yaml
+++ b/tests/overrides.yaml
@@ -2731,14 +2731,25 @@ case_studies/sp500_equity_option_analytics/15_portfolio_management:
   # every label and the case study is long-only, so a scheme survives when k < n_assets.
   # 21 realizes all three; 6 realizes only ew_top5.
   #
-  # Unverified, and stated so it is checked rather than assumed: this case study's model
-  # stages inject max_symbols 5 (06_linear and 07_gbm take the harness default), so if
-  # the sweep ranks predictions those stages produced, its ranked cross-section is 5 and
-  # raising the chain alone realizes nothing - the same shape as us_firm_characteristics
-  # above. Raising them here is not the answer 41 was there:
-  # 11d_stochastic_discount_factor measured 665s at 5 symbols on CI run 34157373964, so
-  # four times the universe is a job of its own. If the PR run shows the sweep still
-  # empty, that is the finding, and it belongs with ml4t/agent-workspace#1084.
+  # **Raising the chain alone realizes nothing here, and that is now measured rather
+  # than predicted.** This case study's model stages inject max_symbols 5 (06_linear and
+  # 07_gbm take the harness default), and the sweep ranks the predictions those stages
+  # produced, so the ranked cross-section stays 5 however wide the price panel is.
+  # 14_backtest fails on all five labels with "top_k_grid[...] = [5, 10, 20] is empty
+  # after filtering against 5 tradeable names (price panel 21, ranked cross-section 5)",
+  # identically on main at 92ce81b2 and on this branch at c14ffeb9. The panel moved; the
+  # cross-section did not.
+  #
+  # So 15, 16 and 17 do not run at all - they fail downstream of 14 - and the 21 here is
+  # a declaration lining the chain up with 14_backtest rather than something CI
+  # exercises. It stays because a chain reading two universes is wrong whether or not
+  # the first stage gets far enough to show it.
+  #
+  # Raising the model stages is not the answer 41 was in us_firm_characteristics:
+  # 11d_stochastic_discount_factor measured 673.0s at 5 symbols on this branch's run
+  # (649.8s on main at the same commit range), which is 43% of the whole job, so four
+  # times the universe is a job of its own. The finding belongs with
+  # ml4t/agent-workspace#1084 and is not chased with a wider number here.
   # The same five labels as 14_backtest above, because this stage takes LABEL
   # too and reads what the previous one registered for that label.
   invocations:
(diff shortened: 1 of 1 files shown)


stefan-jansen:
Both of the two failures that were not shared with `main` came from the base this branch sat on, not from the diff. Rebased onto `92ce81b2`; the rebase was clean, because #831 touched `cme`, `crypto` and `fx` entries in `tests/overrides.yaml` and this branch touches only `sp500_equity_option_analytics` and `us_firm_characteristics`.

The evidence is a same-commit comparison rather than a reading of the diff. `main` at `1bb223bb`, which is exactly what this branch was based on, run [34180296367](https://github.com/stefan-jansen/machine-learning-for-trading/actions/runs/34180296367), failed **ch18-20, cs-crypto_perps_funding, cs-nasdaq100_microstructure and cs-sp500_equity_option_analytics** - the branch's four, with no diff applied. The messages are identical to the branch's, down to the numbers:

| job | notebook | cell | error |
|---|---|---|---|
| ch18-20 | `20_strategy_synthesis/05_portfolio_allocation` | 36 | `OutOfBoundsError: index 0 is out of bounds for sequence of length 0` |
| ch18-20 | `20_strategy_synthesis/06_cost_survival` | 6 | `RuntimeError: No Ch18 cost-sensitivity backtests found for any deployed carrier` |
| cs-crypto | `13_backtest` | 18 | `ValueError: no concentration ... predictions this sweep ranks carry 5 symbols against a price panel of 19` |

Counts match too: ch18-20 `2 failed, 26 passed, 4 skipped` on both; crypto `3 failed, 9 passed, 7 skipped` on both, with 15 and 16 failing downstream of 13 for want of a traded backtest.

**What fixed them was #831**, two commits later. Its own commit message names the crypto failure verbatim and widens the CI fixture to ten perpetuals, because no value in `overrides.yaml` could reach it - the fixture was the five symbols `CRYPTO_SYMBOLS` names. That also settles the reading ml4t/agent-workspace#1089 asks for: the cell-18 concentration refusal is the pre-#831 state, and the cell-14 cadence check that issue is actually about is still gated behind `gpu: true` on `08_tabular_dl`, `09_dl_lstm` and `10_dl_tcn`. Nothing here un-skips them. `main` at `1d3f4807` and at `92ce81b2` is green on both jobs.

So the remaining three - ch11-12, cs-nasdaq100_microstructure, cs-sp500_equity_option_analytics - are `main`'s, at exactly the set #855 sits on.

One thing the new run answers that the old one could not. The `15_portfolio_management` entry carries an explicitly unverified claim: that seoa's model stages inject `max_symbols: 5`, so raising the chain to 21 alone realizes nothing and the sweep stays empty. The previous run could not test it, because the fixture the branch ran against had already been replaced. Against the regenerated one it is a real test, and if the sweep is still empty that is the finding rather than a reason to go looking for another width - it belongs with ml4t/agent-workspace#1084.

stefan-jansen:
The rebased run is in, and both entries that carried a number to be checked were wrong. `6840fa2c` replaces them with what was measured. Comment-only; no declared value changes.

**Every remaining failure is `main`'s, matched at the same commit.** `cs-sp500_equity_option_analytics` and `ch11-12` are error-for-error and count-for-count identical to run [34263597256](https://github.com/stefan-jansen/machine-learning-for-trading/actions/runs/34263597256) on `main` at `92ce81b2` - seoa `16 failed, 19 passed, 2 skipped`, ch11-12 `2 failed, 13 passed, 5 skipped`, same cells, same messages. `ch18-20` and `cs-crypto_perps_funding` are green, as the rebase predicted. All five required checks passed on `c14ffeb9`.

## us_firm_characteristics at 41 costs nothing

The estimate in `05_linear` was +5.5 minutes, derived from `08a_ipca` at 114.9s on 20 symbols scaling with rows. Per-stage, `main` `92ce81b2` against `c14ffeb9`, both `cs-us_firm_characteristics`:

| stage | main | branch | delta | cap |
|---|---|---|---|---|
| `05_linear` + `06_gbm` | 28.4s | 32.0s | +3.6s | 5 -> 41 |
| `11`-`14` chain (10 runs) | 96.1s | 117.4s | +21.3s | 12 -> 41 |
| `08a_ipca` | 124.7s | 75.9s | **-48.8s** | 20 -> 41 |
| whole job | 359.32s | 344.23s | **-15.1s** | |

Every stage other than `08a` moved 0.5-4.4s the *other* way, so the run was not globally faster and `08a` is the only one that moved down. The estimate had the size and the sign of its dominant term wrong. That matters beyond this PR, because ml4t/agent-workspace#1075 prices its 101-symbol option by scaling the same term - that scaling cannot be trusted now. Why IPCA speeds up on twice the cross-section is not established here and I am not guessing at it in a comment.

The fixture claim in that comment is measured rather than asserted now: the fold validation artifacts carry **200 distinct entities** (2200 rows over 11 dates, all three fold directories), on test-data `1ea74a26` - byte-identical to that repo's default branch, which is what CI checks out, since no `ref:` is pinned. So 41 is inside the fold universe and not clipped by it.

## sp500_equity_option_analytics' chain raise realizes nothing

The `15_portfolio_management` entry said this might be the outcome and asked for it to be checked. It is the outcome. `14_backtest` fails on all five labels with

```
top_k_grid['fwd_ret_5d'] = [5, 10, 20] is empty after filtering against 5 tradeable
names (price panel 21, ranked cross-section 5)
```

identically on `main` and here. The model stages inject `max_symbols: 5` and the sweep ranks what they produced, so the price panel moved to 21 and the ranked cross-section stayed at 5.

Two consequences, both written into the entry. `15`, `16` and `17` never run - they fail downstream of `14` - so the 21 on those three is a declaration lining the chain up with `14_backtest`, not something CI exercises. It stays anyway: a chain reading two universes is wrong whether or not its first stage gets far enough to show it. And raising seoa's model stages is not the answer 41 was for firm characteristics - `11d_stochastic_discount_factor` is 673.0s here and 649.8s on `main`, 43% of the job either way. That goes to ml4t/agent-workspace#1084 rather than being chased with a wider number.

Method note, since it is what made the rest of this readable: the durations are differences between consecutive result-line timestamps in the job log, which pytest emits at test completion. They sum to the reported pytest wall time within collection and teardown overhead (-17.4s of stage deltas against -15.1s at the job).

PR #895, opened 2026-09-09
Four infra findings: a render guard, the title docstring, a causal hash, and the placebo universe

stefan-jansen:
Four independent findings from the `lane:infra` queue. None of them obliges a notebook re-run.

## 1. A guard on renders that carry a house-figure warning

Thirteen committed notebooks ship a stderr block under a figure saying the notebook overrode the repository's own figure setup: seven occurrences of `UserWarning: The figure layout has changed to tight`, from a `fig.tight_layout()` over the global `figure.constrained_layout.use: True` in `matplotlibrc`, and twelve of `FigureCanvasAgg is non-interactive`, from a bare `fig.show()` instead of `show_with_alt`. A reader gets a warning about the repository's plumbing where the figure's caption should be, and the count grows with every re-run now that the blanket `warnings.filterwarnings("ignore")` is coming out of the preambles.

The guard names three signatures and holds the thirteen as a debt list that can only shrink, matching the two lists already in `tests/test_notebook_output_hygiene.py`. A notebook leaves the list when it is next re-executed with the offending call removed - that removal is a code-cell edit, so it belongs to whatever pass re-runs the notebook rather than being scheduled on its own.

Deliberately **not** "no stderr in a committed render". Of the 48 notebooks carrying a stderr block, 34 carry a `ConvergenceWarning`, a numerical `RuntimeWarning` or a statsmodels interpolation notice, and those are the one channel a fitted model has to tell a reader it did not converge. A guard that swept them up would push authors back to the blanket filter.

Measured on the base (`237eec23`): 486 committed notebooks scanned, 13 offenders, 19 occurrences. Both directions exercised - dropping an entry fails the offender test, adding a clean path fails the staleness test.

## 2. `add_message_title`'s docstring teaches the current title rule

It still said a title is "a claim, not a label" and that `message` "should state the finding". A figure title now describes what the figure shows, with the interpretation in the markdown around the figure, where correcting it costs no re-run. The function keeps its name: renaming it would edit a code cell in roughly a hundred notebooks.

## 3. A causal hash resolves as a causal run, not as an unknown one

`Result.open` consults `training_runs`, `prediction_sets` and `backtest_runs` and then raises `KeyError: Unknown result hash`. Causal runs are registered into `causal_runs` in the same registry file, so a session checking whether a hash could be migrated was told the row did not exist - about a row sitting in the same database.

It now names the run, names `case_studies.research.causal.CausalResult.open` as the call that reads it, and states the consequence: a causal re-run refits rather than migrates, because `migrate_equivalent_training_identity` reaches training rows only. The check runs after every existing lookup has missed, so the found path pays nothing.

`migrate_equivalent_training_identity`'s docstring now says the same thing, because that is the function a session prices a re-run off, and "a migration script exists" reads as "a version bump is cheap" - which is short by a full refit for whatever part of the work is causal.

Three tests; the first fails against the code it replaces, and the other two hold the boundary (a hash in no table is still unknown; a registry with no `causal_runs` at all still answers rather than raising `OperationalError`).

## 4. The placebo universe is a function of the universe, not of a frame's column order

`run_placebo_benchmark` draws **column positions** - `rng.choice(n_symbols, ...)` then `rets.iloc[:, selected]` - so which symbols a random book holds was decided by the order of the frame it was handed, and nothing said so. Both callers build that frame with `pivot`, which orders columns by first appearance: a property of the parquet the notebook read rather than of any code. Regenerating an upstream artifact in a different row order therefore moved the distribution the strategy is compared against, at the same seed, with the seed making it look pinned.

The sort goes in the helper, so a third caller cannot reintroduce it.

**No reported number moves, verified rather than assumed.** etfs passes sorted columns via a `.sort("symbol", "date")` three lines above its pivot; `us_firm_characteristics` because `labels/prices.parquet` happens to be written in sorted symbol order, checked across its 9,861 symbols. Sorting a sorted list is the identity and `dropna(axis=1, how="all")` preserves relative order. The dependence was real, undeclared, and satisfied by luck at both sites.

Three tests, all three failing against the code they replace. `plot_redundancy_clusters` has the same shape one step weaker and now says so in the source; it is not changed, because all eight callers are closed stage-03 notebooks whose canonical panel pins the draw today.

## Verification

`test_notebook_output_hygiene` 19 passed; `test_placebo_column_order` 3 passed; `test_result_open_names_a_causal_hash` 3 passed; `test_training_identity_migration` 6 passed; `test_causal_result_read` 4 passed. ruff clean.

https://claude.ai/code/session_01WLnJJCNVmBibLcmzZR9cak


stefan-jansen:
Second commit added: `add_message_title`'s docstring still taught the retired rule ("a claim, not a label", "`message` should state the finding"). It now says a title describes what the figure shows, with the interpretation moving to the surrounding markdown where correcting it costs no re-run. Docstring only - no behaviour, and no notebook owes a run for it. The function keeps its name: renaming it would edit a code cell in roughly a hundred notebooks.

stefan-jansen:
Merged `main` in. The only conflict was `utils/style.py`: #898 landed the same `add_message_title` docstring correction while this was open, in a more specific form - it names the quantity, the entities and the axis, where mine said only "describes what the figure shows". Resolved by taking main's side wholesale, so the file is now byte-identical to `origin/main` and that commit contributes nothing to this PR. Four findings, not five.

`ruff check` clean after the merge, which is the check a merge-in needs rather than one before the push: a branch that dropped a then-unused import merges without conflict into a main that has since added a use of that name, and only lint sees it. The 17 `ruff format` suggestions are all committed `.ipynb` on `main` and none is touched here.

PR #917, opened 2026-09-10
Take gpu: true off 25 test overrides that do not need a card, and document the four that keep it

stefan-jansen:
`gpu: true` in `tests/overrides.yaml` means "torch, with a CUDA device it can see"
(`tests/pm_helpers.py:283`), so an entry carrying one is never executed by a runner without a card.
38 entries declared it. This measures every one that can run from the repo venv and removes the flag
where the notebook does not need the hardware.

## Method

Each entry ran through the CI harness with the card hidden, against the CI fixture:

```
ML4T_DATA_PATH=~/ml4t/test-data/data CUDA_VISIBLE_DEVICES="" \
  uv run pytest tests/test_chapter_notebooks.py -k <stem>
```

one at a time, six threads each. `timeout` in `overrides.yaml` is papermill's **per-cell** limit
(`pm_helpers.py:1616` passes it as `execution_timeout`), not a notebook budget, so PASS/FAIL is the
decision variable rather than elapsed-against-budget: `05_synthetic_data/06_llm_tabular_great`
passes at 177s of wall clock against a declared 120s because no single cell takes 120s.

## What changed

**25 flags come off** (38 `gpu: true` entries become 13). 24 on the measurement alone - each notebook ran to completion on CPU with no
cell over its declared limit, which is exactly what its CI job asserts. None of them gets there by
skipping work when no card is visible: every `cuda.is_available()` reference across the 24 is a
device selection, a CUDA-only seeding block, or an `empty_cache()` call.

The 25th, `06_finbert_cross_dataset`, comes off on a reduction. It picks its device with a CPU
fallback at line 102 and has no CUDA guard, so what overran its 300s cell was unbounded work: its
`SAMPLE_SIZE` default of 10,000 exceeds the split's 8,142 rows and never binds, and the entry never
lowered it. At 1,500 it passes in 160s, and the majority-class rate the notebook benchmarks its
accuracy against moves 43.1% -> 42.9%, so the smaller sample does not starve a label class.

**Four keep the flag, and each entry now carries the reason above it** in the form #862 established.

| entry | why |
|---|---|
| `22_rag_financial_research/06_esg_rag_vs_finetune` | `REQUIRE_GPU = True` at line 77 is a module constant, not a papermill parameter; line 83 raises unless `torch.cuda.is_available()`, line 230 raises again if the loaded weights are not on cuda. Fails in 28s with the card hidden. |
| `23_knowledge_graphs/06_gnn_feature_engineering` | Line 73 raises unless `torch.cuda.is_available()`; line 75 pins `DEVICE = "cuda"` for every tensor the fold loop builds at 402-404. Fails in 28s. |
| `10_text_feature_engineering/04_bert_finetuning` | No CUDA requirement in the source - a cost decision. One cell fine-tunes three checkpoints: 637s against a 600s per-cell limit. `MAX_TRAIN_STEPS: 20` already applies; the model count has no parameter. |
| `10_text_feature_engineering/09_filing_text_signals` | No CUDA requirement either. FinBERT chunks each ~6,500-word MD&A section into ~16 windows and costs >15s per filing on CPU, overrunning 600s at both sizes measured. |

The last two say plainly that the flag is about cost rather than hardware, which is a thing a reader
of this file could not previously tell apart from a real requirement.

**Nine entries are untouched** because they cannot run from this venv: `docker_env: py312` (four),
py312 + gensim (one), `ml4t+neo4j+gpu` (two), `requires_import: pfhedge` (one), and one
`tier: on_demand` entry excluded before the GPU guard is reached.

## One removal turned ch10 red, and that is the useful part

`10_text_feature_engineering/07_news_return_signals` also needs no card - it runs on CPU in 28s.
But with the flag off it executed in CI for the first time and **wrote an empty
`news_features.parquet`**, which `08_text_feature_evaluation` then refused with
`ValueError: Empty dataset ... Notebook 07 ran but produced no usable rows`.

The cause is the fixture, not the notebook. Its news sample spans 2020-01-02..2023-12-28 and
`load_us_equities` ends 2018-03-27, so the price join at line 981 drops all 109 feature rows. 07
prints `Overlapping dates: 0` and `Observations: 0` and still exits 0. Production overlaps for nine
years (news 2009-02..2020-06 against prices to 2018-03), so a reader is unaffected.

`tests/fixtures/seed_results.py` had been seeding a synthetic panel for 08 and yields to a real one
with `if path.exists()`; 07 running first overwrote it. That seed's docstring gave 07's `gpu: true`
as its justification, so it is corrected to the real reason and now says what makes it removable.

07 therefore loses the flag and gains `skip` + `skip_reason` naming the fixture gap. It was already
skipping in CI - this makes the stated reason true. Filed as ml4t/agent-workspace#1116, together
with the second half: 07 already computes the overlap and warns on zero, and that branch should
raise rather than write an empty file.

## Two things worth knowing

**A FAIL under load settles nothing, and here it settled the same way anyway.** A run under load
that passes inside its per-cell limit is an upper bound on the same run idle, so only overruns need
a quiet machine. All three were re-measured idle and each came in *faster* than its loaded run by
8-19s (637 vs 645 at load 16.8, 334 vs 353 at load 9.5, 639 vs 649 at load 10.4). Contention was
never the factor.

**Reducing `09_filing_text_signals` is blocked, not merely slow.** Its IC loop needs 20 surviving
observations per signal-horizon pair (line 654); at 12 filings it gets none, and line 721 then
raises `NameError` because the `else` at 707 never binds `ic_summary`. Same displaced line means the
notebook never writes `ic_summary.parquet` on any run, including the shipped one - the committed
`.ipynb` carries the neighbouring save's output and never the IC summary's. Filed separately; a
usable reduction here needs that fixed first.

## Verification

`151 passed` across `tests/test_workspace_prerequisites.py`, `tests/test_invocation_fanout.py`,
`tests/test_pm_helpers.py` and `tests/test_seoa_sweep_universe.py`.

Removing a `gpu` field has no other CI effect: `tests/notebook_catalog.py:239` is its only other
consumer, no workflow or test reads what that produces, and its `GPU_KEYWORDS` set matches the
notebook text anyway.

https://claude.ai/code/session_01WLnJJCNVmBibLcmzZR9cak

### DIFF
--- a/tests/fixtures/seed_results.py
+++ b/tests/fixtures/seed_results.py
@@ -1934,11 +1934,19 @@ def _seed_demo_predictions(cs_dir: Path, cs_id: str, primary_label: str) -> None
 def _seed_news_features(output_dir: Path) -> None:
     """Seed a minimal news_features.parquet for Ch10/08_text_feature_evaluation.
 
-    The panel is produced by 10/07_news_return_signals, which declares `gpu: true` and so
-    skips on a CI runner. 08 is the only chapter-10 notebook that runs there, so without this
-    seed it fails on a missing input rather than exercising anything. The notebook is right to
-    raise on the missing file - a notebook that tolerates absent input reports success for a
-    run that computed nothing - which is why the fixture supplies it instead.
+    The panel is produced by 10/07_news_return_signals, which is marked `skip` because the
+    FNSPID fixture sample (2020-2023) does not overlap the us_equities fixture (ends
+    2018-03-27), so its price join drops every row. 08 is then the only chapter-10 notebook
+    that runs on a CI runner, and without this seed it fails on a missing input rather than
+    exercising anything. The notebook is right to raise on the missing file - a notebook that
+    tolerates absent input reports success for a run that computed nothing - which is why the
+    fixture supplies it instead.
+
+    This seed cited 07's `gpu: true` until 2026-09-10. That flag was wrong: 07 runs to
+    completion on CPU in 28s. Removing it made 07 execute ahead of 08 and overwrite this panel
+    with the empty one its own join produces, which is how the real reason surfaced. Once the
+    fixture's news and price windows overlap (ml4t/agent-workspace#1116), 07 runs, writes a
+    real panel, and this seed can go - `if path.exists()` already yields to it.
 
     The notebook loads from get_output_dir(10, "fnspid") / "news_features.parquet"; in test
     mode that is {ML4T_OUTPUT_DIR}/ch10_fnspid/news_features.parquet. This directory name is
--- a/tests/overrides.yaml
+++ b/tests/overrides.yaml
@@ -383,9 +383,22 @@
     SAMPLE_SIZE: 1500
   timeout: 300
 10_text_feature_engineering/07_news_return_signals:
+  # Not a GPU notebook - it ran to completion on CPU in 28s with the card hidden. What it cannot
+  # do is produce a row against this fixture: the news sample spans 2020-01-02..2023-12-28 and
+  # `load_us_equities` ends 2018-03-27, so the join at line 981 drops all 109 feature rows and
+  # the notebook writes an empty news_features.parquet. It reports that ("Overlapping dates: 0",
+  # "Observations: 0") and still exits 0, so with the GPU flag off it silently replaced the panel
+  # `tests/fixtures/seed_results.py::_seed_news_features` puts there for 08, which then failed on
+  # the empty file. Production overlaps for nine years (news 2009-02..2020-06, prices to
+  # 2018-03), so this is the fixture's sampling window, not the notebook.
+  # ml4t/agent-workspace#1116.
   parameters:
     MAX_ARTICLES: 500
     MAX_TICKERS: 5
+  skip: true
+  skip_reason: >-
+    The FNSPID fixture sample (2020-2023) has no overlap with the us_equities fixture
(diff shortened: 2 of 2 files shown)


stefan-jansen:
Heads-up on a cross-PR interaction that will not show up as a conflict.

I test-merged this branch's `tests/overrides.yaml` against #911's (chapter 22) with `git merge-file`: rc=0, zero conflicts, and the merged file parses at 471 entries with both changes intact - this PR's `gpu: true` count still goes 42 to 17, and #911's rewrite of `01_sec_filing_pipeline` survives. So the text merges fine in either order and nothing here needs rebasing on that account.

What does not survive is the comment this PR adds above `22_rag_financial_research/06_esg_rag_vs_finetune`. It cites three line numbers:

> `REQUIRE_GPU = True` at line 77 [...] line 83 raises unless `torch.cuda.is_available()`. Line 230 raises again if the loaded FinBERT parameters are not on cuda.

#911 rewrites that notebook and moves all three. On its branch they are 86, 92 and 322. The substance stays true - `REQUIRE_GPU = True` and both raises are still there, so the entry is right to keep `gpu: true` - but if #911 merges after this one, the justification points at three lines that say something else, and nothing will flag it.

Two ways out, both cheap. Cite the symbols instead of the line numbers (`REQUIRE_GPU`, the `torch.cuda.is_available()` guard, and the `parameter_device.type != "cuda"` check in the fine-tuning path), which is stable across either merge order. Or merge this one second and refresh the numbers. The first is better, because the same comment has to survive the next edit to that notebook too.

I have not touched either branch.

PR #964, opened 2026-09-11
12/08 can be reduced, and the loader can be asked for columns

stefan-jansen:
Four changes, three of them in the same family: a parameter that is declared and reaches nothing.

**12/08's `MAX_SYMBOLS` now reaches the data it names.** The notebook declared it and called `load_modeling_dataset` without it at both sites, so CI ran the full ETF universe and then the full universe of two more case studies for the publication beeswarm, inside a 300s budget. Both calls take the cap, and the override sets it under `parameters:` where papermill can see it rather than under `env:`, which the chapter test never passes on. Re-rendered at production settings; outputs are unchanged except the printed artifact path, which names the worktree that ran it. The reduced run measures 72s on four cores against the 300s budget kept here.

**`load_modeling_dataset` gets a `columns` parameter, pushed into the scans.** Every caller materialized the whole joined panel and projected afterwards. On `us_equities_panel` with label `fwd_ret_1d` the join returns 74 columns and 5.88 GB, the DML estimand reads 7 of them worth 0.47 GB, and the notebook peaks at 17.6 GiB doing it; `max_symbols` cannot help, because it narrows rows and a caller that needs the whole universe cannot use it at all. The request is unioned with the join keys, the detected date and entity columns and the label rather than taken literally, which is why the parameter belongs in the loader. Passing nothing keeps today's behaviour.

The test worth reading is the third of the six. A projection applied to the returned frame passes "the right columns came back" and "an unknown name raises" exactly as well as one pushed into the scan, and buys none of the memory, so the probe records every column set that reaches a `collect()` and asserts none of them held a dropped column - with the unprojected load as the control proving the probe can see one. Mutation-tested: moving the projection after the join leaves five of the six green and fails that one.

**`DISABLE_HPO` is deleted.** The harness put it in every notebook's environment beside `MPLBACKEND` and the thread caps; one occurrence in the repo, the setter, unread since it was introduced on 2026-06-22. Deleted rather than wired up, because reductions here arrive as `parameters:` in `tests/overrides.yaml` and an env var a notebook reads for itself would be a second reduction channel invisible where reductions are declared.

**One unrelated test, repaired rather than skipped.** `test_sp500_options_corpus_uses_financial_timeline_geometry` groups the model-based artifact by `fold`, and sp500_options writes a fold-free artifact now, so it raised wherever the canonical artifact is reachable and passed only where it is absent. Found by linking the sp500_options artifacts into this worktree so chapter 12 could build both panels of its figure - the test had been skipping here for want of the artifact, not passing.

Skipping it outright would have ended the check rather than narrowed it, since it covers one case study. Its two assertions do not need the same thing: the claim it is named for, that the corpus resolves boundaries against the financial timeline rather than the label one, needs no `fold` column and now runs unconditionally - resolved train_starts are 2017-02-02 and 2018-01-05 against 2017-01-05 and 2018-01-04 derived from the labels. Only the per-fold start comparison is guarded, and the guard names what covers the fold-free write instead: `04_model_based_features.py:411` and `:1470`, where every block's `fit_end` is required to precede the session it speaks for.

Not included: wiring `columns=` into `case_studies/utils/causal.py`. Its column list is built from a config resolved after the load, so passing it needs a reordering of that function, and the memory win there is the case study's to claim.

https://claude.ai/code/session_01VEtXvNk14uZckwswthVihc

### DIFF
--- a/12_gradient_boosting/08_shap_analysis.ipynb
+++ b/12_gradient_boosting/08_shap_analysis.ipynb
@@ -2,13 +2,13 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "91a20e01",
+   "id": "da72904e",
    "metadata": {
     "papermill": {
-     "duration": 0.002132,
-     "end_time": "2026-09-10T08:15:04.604803+00:00",
+     "duration": 0.006197,
+     "end_time": "2026-09-11T14:05:49.389792+00:00",
      "exception": false,
-     "start_time": "2026-09-10T08:15:04.602671+00:00",
+     "start_time": "2026-09-11T14:05:49.383595+00:00",
      "status": "completed"
     }
    },
@@ -44,13 +44,13 @@
   },
   {
    "cell_type": "markdown",
-   "id": "1f45a2f9",
+   "id": "5dd2bde7",
    "metadata": {
     "papermill": {
-     "duration": 0.001503,
-     "end_time": "2026-09-10T08:15:04.608100+00:00",
+     "duration": 0.001661,
+     "end_time": "2026-09-11T14:05:49.394658+00:00",
      "exception": false,
-     "start_time": "2026-09-10T08:15:04.606597+00:00",
+     "start_time": "2026-09-11T14:05:49.392997+00:00",
      "status": "completed"
     }
    },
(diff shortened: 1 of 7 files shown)


stefan-jansen:
`test-unit-data` here is red for a reason outside this PR, and a re-run will not
clear it: rebase onto `main` at `a8af4439` or later.

`test_the_manifest_subsets_are_exactly_the_declared_ones` is the only failure
(1 failed, 110 passed). It compares the fixture manifest's subsets against
`tests/create_test_data.py`'s declarations in both directions, and the fixture repo
is unpinned - every branch reads its `main`. I pushed a manifest carrying three new
SEC subsets shortly before this run started, and the public half that declares them
landed as #963 at 14:31, four minutes after this job began. Between those two
moments any branch without the declarations fails that one test, and this branch
was in that gap.

My sequencing error, not anything here. Sorry for the noise.

PR #969, opened 2026-09-11
etfs/15_portfolio_management states its reuse without opening on a zero

stefan-jansen:
## The disclosure

On a warm registry the sweep printed `0 computed, 180 served from the registry`. A reader of
the chapter rather than of the run takes that first clause as a sweep that computed nothing;
it reused 180, which is the intended idempotent behaviour and the opposite of doing nothing.
The disclosure itself has to stay - a notebook that serves results from the registry without
saying so is a silent skip - so this is a wording change.

`reuse_disclosure` in `case_studies/research/execution.py` already words all four cases and is
what `crypto_perps_funding`'s 14, 15 and 16 call. This routes `etfs` through it rather than
spelling a fifth phrasing. Re-executed; the committed output now reads:

    Sweep complete in 0.2 minutes: reused all 180 from the registry, computed by an earlier run

## Three sites deliberately not in this PR

`crypto_perps_funding/13_backtest`, `fx_pairs/13_backtest` and
`fx_pairs/14_portfolio_management` carry the same count-pair wording. The edit is written and
held back because none of the three can be re-executed today: each stops on a
population-supersedes guard wanting an explicit declaration about production data
(`a changed population named '...' must explicitly supersedes ...`). Making that declaration
is a decision about what the canonical population is, not something to do as a side effect of
changing printed text. They ride along with their next re-run.

`crypto_perps_funding/13_backtest` would in fact commit without complaint, because it carries
no provenance stamp and the gate cannot compute staleness for it - which is a reason to be
careful with it rather than a licence.

## The alt text in the same file

The second commit fixes the allocator chart's alt text, in the notebook this branch already
re-executes. It claimed "every allocation-stage backtest" where the frame is scoped to
`ADVANCED_PREDICTIONS`, and it ended on a computed count and value range that N8 bans. Caught
by nbcheck's universal-word advisory; settled by reading the frame.

## Based on

`fix/backtest-explorer-null-metric-columns`. Without it `etfs/15_portfolio_management` cannot
execute, so this PR's notebook could not be refreshed.

### DIFF
--- a/case_studies/etfs/15_portfolio_management.ipynb
+++ b/case_studies/etfs/15_portfolio_management.ipynb
@@ -2,13 +2,13 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "89b1d269",
+   "id": "dcc050a9",
    "metadata": {
     "papermill": {
-     "duration": 0.00227,
-     "end_time": "2026-09-11T14:35:00.466978+00:00",
+     "duration": 0.007499,
+     "end_time": "2026-09-11T14:39:02.860654+00:00",
      "exception": false,
-     "start_time": "2026-09-11T14:35:00.464708+00:00",
+     "start_time": "2026-09-11T14:39:02.853155+00:00",
      "status": "completed"
     }
    },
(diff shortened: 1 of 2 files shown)


stefan-jansen:
This PR runs **zero** checks, and that is structural rather than slow: `.github/workflows/test.yml`
is `on: pull_request: branches: [main]`, so a PR based on another branch fires no workflow.
`check-runs` for `4f7f5e9c` returns `total_count: 0`; the five required contexts will never appear.

So this cannot be merged from here. Once #968 is on `main` I will rebuild these three commits on
top of it and open the replacement against `main` - a rebuild rather than a retarget, because #968
squash-merges and its commit will not be an ancestor of `main` afterwards. This PR will be closed
pointing at that one.

stefan-jansen:
Superseded, not abandoned. #968 is on `main` as `496ea37c`, so these three commits are rebuilt on
top of it and reopened against `main`, where CI actually runs. Replacement: https://github.com/stefan-jansen/machine-learning-for-trading/pull/971

Closing this one because GitHub can PATCH a PR's base but never its head, and keeping this number
would have meant force-pushing the branch.

PR #970, opened 2026-09-11
Make a skip declare the condition its reason rests on, and check it

stefan-jansen:
`skip: true` in `tests/overrides.yaml` takes a notebook out of CI on a written reason, and
nothing asked whether the reason was still true. The suite was green whether it was sound,
stale, or never true, so a wrong reason was invisible by construction and the notebook behind
it stayed out of CI indefinitely.

Every skip now declares the condition beside the prose:

```yaml
case_studies/etfs/18_holdout_predictions:
  skip: true
  skip_reason: "..."
  skip_blocker:
    no_canonical_selection:
      of: etfs
```

`tests/skip_blockers.py` returns a reason while the condition holds and `None` once it has
expired. The notebook tests honour the skip only while a reason comes back, so a fixture that
grows the file or a registry that gains a selectable candidate retires the skip with no edit to
any file, and `tests/test_skip_blockers.py` fails the build on a declaration that has outlived
its reason.

This is `tests/awaiting_rebuild.py`'s mechanism pointed at a different registry. That one
measures against `ML4T_ARTIFACT_ROOT`, the maintainer's own artifacts; `overrides.yaml` records
at `case_studies/sp500_options/11_model_analysis` why that is the wrong instrument for a skip
about the CI fixture, because such a condition reads as satisfied on any workstation that has
run the models while the fixture the job runs against still holds nothing. Everything here is
measured against the fixture: `ML4T_DATA_PATH` for files, the seeded `ML4T_OUTPUT_DIR` for
registries.

Four kinds are decidable (`absent_fixture_path`, which takes a glob, `no_canonical_selection`,
`absent_population`, `absent_candidate_set`). Two say they are not rather than pretending:
`external`, for a service, credential or device the runner does not have, which must sit off
the per-commit tier, and `fixture_shortfall`, which must carry the date its reason was last
confirmed by running the notebook. A ceiling keeps those two from becoming the comfortable
answer.

## What the check found while it was being written

Four reasons were false, and the mechanism is what surfaced them.

| row | the reason said | what is true |
|---|---|---|
| `15_causal_estimation/10_case_study_insights` | CI has none of the nine case-study registries | the seeded output dir carries all nine and the notebook passes in 68s. **Skip removed** |
| `24_autonomous_agents/10_framework_comparison` | langgraph is not in the CI image | it is a main dependency, placed there so it reaches the image; the notebook passes on the weekly tier. **Skip removed** |
| `02_financial_data_universe/06_futures_continuous` | test data has only continuous series | `futures/market/individual` ships CL, ES and NQ. It fails on a missing `futures/market/contract_definitions.parquet`, which is now what the reason says and what the blocker checks |
| six rows in etfs, nasdaq100_microstructure, sp500_options | the seeded training specs carry no `computation` block | they carry one - 136/140, 97/97, 106/106. What stops them is that every ranked validation backtest belongs to a superseded generation |

Reasons were also rewritten for cme_futures 17 and 18 ("registers no validation backtest" - it
registers 68), us_firm_characteristics 15 ("holds no validation backtests" - it holds 75),
fx_pairs 17, etfs 20 and cme_futures 10b, each against what running it or querying the fixture
registry actually produced.

## The guard runs in a job that has the fixture

The first version of it asserted nothing: its per-row checks measure against the fixture, and
`test-unit`, the only job collecting the file, checks out no test data, so `populated_data_dir`
skipped every one. `test-unit-data` exists for exactly that failure, checks out the test-data
repo, and fails when any test in it skips. The file is in its list, and the undecidable kinds
no longer report as skips.

Verified by reproducing that job rather than reasoning about it: a venv built from its own
package list under `uv export --frozen --extra live --extra dev` as a constraint,
`ML4T_OUTPUT_DIR` unset the way the step unsets it, run against the test-data checkout. **75
passed, 0 skipped**, and the job's junit-XML skip gate reads PASS.

Both evaluator directions are exercised on real fixture data, and the branch the fixture cannot
reach today - a resolver that returns - is covered so it is not dead code.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1027,7 +1027,13 @@ jobs:
           # whether that file reads a redirect or the source tree. Both orders were
           # measured before this change landed and both are green: the roster tests
           # read a SEEDED redirect, which is the case the unset above never covered.
+          # tests/test_skip_blockers.py belongs here and in no other job. Its expiry checks
+          # measure a declared skip condition against the fixture - files under
+          # ML4T_DATA_PATH and the registries seeded_output_dir builds - so in test-unit,
+          # which checks out no test data, every one of them skips and the guard asserts
+          # nothing. That is the same "runs in no job at all" state this job was created for.
           .venv/bin/python -m pytest \
+            tests/test_skip_blockers.py \
             tests/test_eoa_universe_roster.py \
             tests/test_fixture_registry_identity.py \
             tests/test_reduced_universe_is_shared.py \
--- a/tests/overrides.yaml
+++ b/tests/overrides.yaml
@@ -25,8 +25,10 @@
   timeout: 300
 02_financial_data_universe/06_futures_continuous:
   skip: true
-  skip_reason: "Requires individual contract months \u2014 test data has only continuous series"
+  skip_reason: "Reads `futures/market/contract_definitions.parquet`, which the CI fixture does not carry: cell 11 raises FileNotFoundError on it. Measured 2026-09-11 by un-skipping and running it. The individual contract months the older reason said were missing are present - `futures/market/individual` ships CL, ES and NQ - so that reason had stopped describing the fixture."
   timeout: 300
+  skip_blocker:
+    absent_fixture_path: futures/market/contract_definitions.parquet
 02_financial_data_universe/07_sp500_options_eda:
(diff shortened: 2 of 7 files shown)


stefan-jansen:

PR #972, opened 2026-09-11
A column projection must not narrow the identity the run records

stefan-jansen:
Based on `teaching/ch-12-shap-reduction`, which adds `columns=` to `load_modeling_dataset`. That branch deliberately left `case_studies/utils/causal.py` alone, because it resolves its column list from `setup.yaml` *after* the load, so using the parameter needs the function reordered. This is that change, plus the reason it could not be a one-line wiring.

## The projection re-keys the run

`columns=` pushes a projection into the parquet scans. It changes no rows and no retained value, so it reads as a pure memory optimization. It is not one for any caller that registers a run.

`feature_names` is derived from the joined frame (`utils/modeling.py:1005`), so a projection narrows it, and it is hashed into every registered run twice - as `computation.feature_names` and again inside `computation.input_data_spec` - both reached by `training_hash_from_spec`. Wiring the projection in as-is would have re-keyed **twenty registered `causal_runs` across seven case studies**, for a change that reads the same artifacts, keeps the same values in the columns it retains, and computes the same estimate. The lane doing the wiring is not the lane that pays: `us_equities_panel` holds none of the twenty.

`build_modeling_input_lineage` already ruled on this shape in the other direction, in its note on why `feature_dtype` is written only when it is not the default: a key that moves the fingerprint of a case study whose declaration did not change invalidates every run registered against it, and "any future default must be added the same way, for the same reason". A projection narrowing the recorded list is the same event.

## What changed

- **`ModelingDataset.panel_feature_names`** - the feature list the artifacts carry, independent of what one caller chose to read. `input_lineage` and the causal spec read it; `feature_names` still describes the frame the caller was handed, which is what its other consumers want. Defaults empty, so hand-built `ModelingDataset` objects behave exactly as before.
- It is reconstructed from the column lists captured *before* the projection. The one shape that reconstruction cannot express - a name carried by both the financial and the model-based artifact, which polars would suffix `_t` at a position not recoverable from the source lists - raises rather than guesses. No case study has one; checked across all eight.
- **`resolve_causal_request` moves four lines.** `setup`, `causal`, `treatment` and `confounders` are read before the load rather than after it, still behind `study.activate(tier)` because that resolves the `study.root` they read. Nothing in that block depends on the panel, which is why the reordering is available at all. The loader adds the date, entity and label columns itself, so the request is just the treatment and the confounders.

## Measured

On `us_equities_panel/fwd_ret_1d`, 2026-09-11: the projected load hands back 9,973,233 x 7 at 0.441 GiB for a process peak of **4.93 GiB**. The unprojected join is 74 columns whose numeric block alone is 5.39 GiB - that second figure is arithmetic from the three artifact schemas, **not a measurement**, because there was no headroom to load it beside a running notebook. The full-scale cost this targets is 17.6 GiB on `14_causal_dml`.

## Verification

The identity claim does not rest on the unit tests. Against **four registered production runs** - three `fx_pairs` labels and `us_firm_characteristics`, which has no model-based artifact at all - a projected load reproduces the stored `spec_json`'s `computation.feature_names`, its `input_data_spec.feature_names` and its fingerprint exactly, while narrowing the load to 7 columns of 63 and 57 features respectively. The same check confirms each stored spec rehashes to its registered `causal_hash`, so the comparison is against the registry's own hashing rather than a reimplementation of it.

Six new tests, with the control that matters last: the fingerprint equality would be satisfied by an implementation that never projects, so one test narrows the recorded list by hand and asserts the fingerprint *does* move. Mutation-tested - pointing `input_lineage` back at `feature_names` fails the fingerprint test with the narrowed list in the assertion.

| suite | result |
|---|---|
| projection + identity (new file) | 12 passed |
| causal resolver | 63 passed |
| loader / sidecar / storage-dtype / fold coverage | 58 passed |
| linear resolver | 77 passed |
| folds, CV windows, artifact specs, model planning | 52 passed |

`tests/test_causal_adapter.py` fakes the modeling dataset with three `SimpleNamespace` doubles that did not carry the new field, so the resolver raised `AttributeError` against them. The doubles are fixed rather than the production read defended with `getattr`: a double missing an attribute the code under test reads is the defect, and a default would have made the next missing field silent.

## Merge order

Stacked on `teaching/ch-12-shap-reduction`. CI workflows filter on `base: main`, so this PR will show **zero checks** until its base merges - that is unmergeable rather than pending. Merge the chapter-12 PR first, then this one retargets to `main` and gets a real run.

### DIFF
--- a/case_studies/utils/causal.py
+++ b/case_studies/utils/causal.py
@@ -1366,10 +1366,25 @@ def resolve_causal_request(study: Study, request: dict[str, Any]):
     study.require_writable()
     study.activate(tier)
     label_ref = study.labels.get(request["label"], execution_tier=tier)
+    # Read before the load, not after, which is the whole of the reordering. The estimand needs
+    # exactly the treatment, the confounders, the outcome and the two join keys; the loader adds
+    # the last three itself, so the only names supplied here come from setup.yaml and nothing in
+    # this block depends on the panel. Loading first and projecting the finished frame - which is
+    # what this did - has already paid for every column. Measured on us_equities_panel/fwd_ret_1d,
+    # 2026-09-11: the projected load hands back 9,973,233 x 7 at 0.441 GiB for a process peak of
+    # 4.93 GiB, where the unprojected join is 74 columns whose numeric block alone is 5.39 GiB
+    # (computed from the three artifact schemas, not loaded - there was no headroom to load it
+    # beside a running notebook, so that figure is arithmetic and the two beside it are measured).
+    # `study.activate` still precedes both, because it resolves the `study.root` read below.
+    setup = yaml.safe_load((study.root / "config" / "setup.yaml").read_text()) or {}
+    causal = setup.get("causal") or {}
+    treatment = str(causal["treatment"])
+    confounders = tuple(str(value) for value in causal["confounders"])
     mds = load_modeling_dataset(
         study.case_study,
         label_ref.name,
         max_symbols=int(reductions.get("max_symbols", 0)),
+        columns=[treatment, *confounders],
     )
     if mds.date_col != "timestamp" or not mds.entity_cols:
         raise ValueError("DML runner requires timestamp and an entity key")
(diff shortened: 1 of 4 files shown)


stefan-jansen:

stefan-jansen:
Superseded by #984. `teaching/ch-12-shap-reduction` squash-merged as `625b707f`, so its eight commits are not ancestors of `main` and this PR's base cannot be retargeted without rewriting history. #984 is the same single commit cherry-picked onto `main` - byte-identical patch - against a base that gets the full CI run this one could not (workflows filter on `base: main`, so only the four platform checks ever ran here).

PR #973, opened 2026-09-11
Ask the input-artifact vintage question before the chain is queued

stefan-jansen:
A kept-registry case study has four pre-flight resolutions that can refuse a run: input
artifact vintage, population hash, causal identity, and candidate sets.
`scripts/check_supersedes_literals.py` covers the last three. The first had no warning but the
run itself stopping.

`_enforce_input_artifact_vintage` refuses a run that would join a population fitted on a
different vintage of the same input artifact. Regenerate a stage-03 or stage-04 artifact and
every later run refuses until `declare_artifact_supersession` records which sha the new one
replaces. On 2026-09-10 `us_equities_panel/06_linear` hit exactly that at plan time:
`model_based.parquet` on disk hashed `0e74d15f` while the registered training runs for
`fwd_ret_1d` had been fitted against `86ece972`.

`scripts/check_input_artifact_vintage.py` asks the same question ahead of the launch. It is
its own script rather than a fifth section of the literals checker because what it costs
differs: this refusal fires *before* the fit, so it costs a launch and a queue slot rather
than compute, and conflating that with resolutions that cost a whole fit loses the
distinction.

The comparison is now `input_artifact_vintage_conflicts`, extracted from the enforcement and
returning values instead of raising. The enforcement raises from it and the script calls it. A
pre-flight that re-implemented the rule could pass where registration refuses, which is a
green light for a chain that will not run.

## What the check has to get right

A declared supersession is what makes a second vintage legitimate, so a check reading only the
disk and the runs would report every deliberate regeneration as a defect, and would be red on
`us_equities_panel` today. The condition is narrower: the sha on disk differs from what the
runs pin **and** `artifact_supersessions` holds no row retiring the pinned one.

What it does not look at, it says. Two things fall outside:

- An artifact name it cannot locate. `financial`, `model_based` and `label` resolve from the case study's own specs, `eval_label` through the classification mapping the loader uses; anything else is reported `unresolved`.
- A run whose inputs are recorded in a shape the rule does not read. Measured across the nine live registries: **121 runs** - 45 `latent_factors` recording `input_data_spec.files` and 76 `deep_learning` nesting the payload at `input_data_spec.input_data_spec.artifacts`. Neither reaches `_input_artifact_shas`, so **registration does not vintage-check them either**. They are reported `unchecked` per family rather than read: a pre-flight stricter than the rule it previews would refuse a chain that would in fact register. That enforcement gap is a separate defect and is filed as one.

The four non-conflict states are distinguished rather than lumped under one label, because
calling a declared replacement `current` claimed runs had been fitted on a file no run had
touched, and `--quiet` then hid it: `current`, `superseded`, `mixed`, `accepted_replacement`.

## Verification

Against the nine live registries: **95 pairs - 83 current, 1 superseded, 11 unchecked, 0
undeclared, 0 unresolved, 0 absent, exit 0.** The superseded pair is
`us_equities_panel/fwd_ret_1d/model_based`, and it is superseded because a declaration retires
`86ece972`, not because two shas are pinned.

Seven tests cover both directions on a constructed registry, with the two unread spec shapes
copied from the live registries rather than invented. 34 tests green across this file,
`tests/test_input_artifact_vintage.py` and `tests/test_artifact_supersession.py`.

### DIFF
--- a/case_studies/utils/registry/registration.py
+++ b/case_studies/utils/registry/registration.py
@@ -8,6 +8,7 @@ import os
 import shutil
 import sqlite3
 import uuid
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


stefan-jansen:

PR #974, opened 2026-09-11
Break the reported winner's tie on a name, not on row order

stefan-jansen:
Two rankings in `build_all_synthesis` sort on one key and take `head(1)`, and each reports a
NAME: the best allocator and the best risk overlay. A tie goes to whichever row arrived first,
so the allocator a case study reports is not a function of its data. It is the shape where a
visible precaution - the sort - pins one dimension and leaves another free.

## The tie is measured, not hypothetical

Across the nine live registries on 2026-09-11, **14 (case study, stage) pairs hold at least one
exactly repeated Sharpe**, with multiplicity up to 6, and `crypto_perps_funding`'s signal stage
holds two rows tied *at the maximum*. Three of the pairs are `risk_overlay` stages, which is
the frame the second site ranks. The mechanism is ordinary: two overlays that never trigger
book the baseline exactly.

Both sites now call `rank_one`, which sorts on the value and then the name. The second key is
consulted only where the first is equal, so nothing moves where there is no tie.

`nulls_last=True` came out of writing the test: polars sorts nulls **first** under
`descending=True`, so a row with no Sharpe would have won the ranking and been reported as the
best of them. No live registry holds a null Sharpe today, so this repairs no current number;
the frame is built by a join and the code reading the result already guards a null
`max_drawdown`.

## Verification

Four tests, and the control is the load-bearing one: the same permutations of the same tied
frame run against the one-key sort this replaced, asserting it returns two different winners.
Asserting the fix alone would pass equally against a frame that never ties, and the point is
that this frame does.

12 tests green with `tests/test_holdout_selection_is_single_sourced.py`; 121 passed and 8
skipped across the wider strategy-analysis selection.

The three instances that motivated the issue were re-verified as already fixed before anything
was touched: the seeded `hmmlearn` fit runs inside `threadpool_limits(1)`, the bootstrap
interval sorts its input before `rng.choice`, and the unpinned evidence script is gone. The
rest of the sweep is adjudicated and filed for the bands that own those chapters.

### DIFF
--- a/case_studies/utils/strategy_analysis.py
+++ b/case_studies/utils/strategy_analysis.py
@@ -2395,6 +2395,31 @@ def classify_holdout_degradation(
     return "degenerate"
 
 
+def rank_one(frame: pl.DataFrame, *, by: str, name: str) -> pl.DataFrame:
+    """The top row of *frame* by *by*, with *name* deciding a tie.
+
+    A one-key ``sort(by, descending=True).head(1)`` hands a tie to whatever order the frame
+    arrived in, so the row it returns is not a function of the data - the defect class
+    ml4t/agent-workspace#333 is about, where a visible precaution (the sort) leaves one
+    dimension of the answer free.
+
+    Exact ties in these quantities are not hypothetical. Measured across the nine live
+    registries on 2026-09-11, 14 (case study, stage) pairs hold at least one exactly repeated
+    Sharpe, with multiplicity up to 6, and crypto_perps_funding's signal stage holds two rows
+    tied at the maximum. Two risk overlays that never trigger book the baseline exactly, which
+    is the mechanism that produces them.
+
+    Breaking on the name costs nothing when there is no tie: the second key is only consulted
+    where the first is equal.
+
+    ``nulls_last`` because polars puts nulls FIRST under ``descending=True``, so a row with no
+    value for *by* would win the ranking and be reported as the best of them. No live registry
+    holds a null Sharpe today, so this changes no current result; it is here because the frame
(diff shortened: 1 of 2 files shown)


stefan-jansen:
Checked whether this moves any result that is currently published, since a tie-break only matters where the top ties.

Across the nine live registries, the maximum Sharpe in every `allocation` and `risk_overlay` stage is held by exactly one row:

| case study | stage | rows | rows at the max |
|---|---|---|---|
| cme_futures | allocation / risk_overlay | 120 / 28 | 1 / 1 |
| crypto_perps_funding | allocation / risk_overlay | 720 / 56 | 1 / 1 |
| etfs | allocation / risk_overlay | 378 / 60 | 1 / 1 |
| fx_pairs | allocation / risk_overlay | 180 / 84 | 1 / 1 |
| sp500_equity_option_analytics | allocation / risk_overlay | 900 / 70 | 1 / 1 |
| sp500_options | allocation | 60 | 1 |
| us_firm_characteristics | allocation | 240 | 1 |

So no reported allocator or overlay changes and nothing needs re-rendering. The ties that exist sit below the maximum, except in `crypto_perps_funding`'s signal stage, which neither of these two sites ranks.

One limit worth stating rather than glossing: these are whole-stage maxima, and both call sites rank a frame already filtered to a carrier prediction. A subset maximum could tie where the stage maximum does not, so this bounds the blast radius rather than proving it empty. The tie-break is correct either way - it only decides a case that is otherwise decided by row order.

stefan-jansen:

PR #975, opened 2026-09-11
An int constant subscripting a container built from another constant now fails the gate, not the reader

stefan-jansen:
An integer constant used as a bare subscript into a container built from a *second*
constant is a coupling with nothing checking the two agree. Change either and the notebook
raises `KeyError` the next time it executes, while ruff, the formatter, `py_compile`, the
prose checker and `tests/overrides.yaml` all stay green. Four review rounds produced five
execution failures of this shape, none of them catchable by anything CI ran.

## The checker, and why it is one rule rather than four fixes

`.github/scripts/check_constant_subscripts.py`, wired into the `guards` job. Stdlib AST,
nothing imported or executed.

The filter is the rule: a **module-level integer** constant used as a **bare positional
subscript** into a container this module builds from constants. Both halves earn their
place, measured rather than asserted. Dropping the integer half admits string keys into
dicts a reader never edits. Dropping "container this module builds" takes the sweep from 6
hits to 18, and every one of the extra twelve is `array[idx]` with `idx = 0` into a frame
whose length the data decides - no second constant, nothing to disagree with. I checked all
twelve rather than assume, which is how chapter 1 turned up.

Four container shapes reach the real hits, and the last two were necessary rather than
thorough:

- a dict or list literal;
- a dict filled in a loop over a constant list;
- a frame whose index is that loop's constant list, `pd.DataFrame(rows).set_index("block_size")`.
  Without this the rule reports the two mild cases and **misses the only reader-reachable
  one**, chapter 15's, which is the hit the issue leads with;
- a helper that fills a local dict in a loop over one of its parameters and returns it.
  This is how `01_process_is_edge/factor_regimes.py:550` turned up, a genuine hit that was
  **not in the issue** because the sweep that filed it matched only literals and in-line loops.

A second rule covers the wrong fix rather than the defect. Deriving one constant from the
other inside the papermill `parameters` cell looks like it removes the coupling and is
worse: injection happens *after* that cell, so the derived name keeps its default when the
list is overridden, and a loud `KeyError` becomes a result quietly drawn from a
configuration that never ran.

`--selftest` runs 26 constructed cases, 11 that must fire and 15 that must not, and CI runs
it before the sweep: a rule not shown to fail on a constructed case is not worth gating on.
Two of those cases caught defects in this checker. An empty `rows = []` was being claimed by
the int-sequence branch, so the must-fire DataFrame case reported nothing while both of its
must-not-fire siblings passed vacuously - an absence presenting as success inside the
instrument built to detect one.

## The four notebooks

Each gets a guard where the constants are declared, or - for chapter 15, whose constants are
papermill parameters - in the first cell after the injection point, so an override is covered
and the failure arrives before the expensive sweep rather than after it.

| notebook | coupling |
|---|---|
| `01_process_is_edge/factor_regimes.py:550` | `gmm_grid[N_REGIMES_SELECTED]` against `N_REGIMES_GRID` |
| `09_model_based_features/11_hmm_regimes.py:529` | `fitted_models[N_STATES]` against `STATE_COUNTS` |
| `10_text_feature_engineering/08_text_feature_evaluation.py:389,439,466` | `RET_COL_BY_HORIZON[KEY_HORIZON]` |
| `10_text_feature_engineering/08_text_feature_evaluation.py:351` | `RET_COL_BY_HORIZON[h]` for every `h` in `CONFIG.horizons`, declared 119 lines apart - more reachable than the one filed, and not in the issue |
| `15_causal_estimation/04_dml_crypto_regime.py:737` | `block_sweep_df.loc[BLOCK_SIZE_HEADLINE]` against `BLOCK_SIZES`, both in the parameters cell |

## What the re-renders show

Chapters 1, 9 and 15 are output-neutral, verified cell by cell against the committed render
rather than claimed. Chapter 9 is byte-identical across all 23 code cells. Chapter 1's only
difference is the memory address inside a pandas Styler's `text/plain` repr, six cells of
`<pandas.io.formats.style.Styler at 0x...>`, which is per-render noise already sitting in the
committed file.

Chapter 10's numbers moved, and not because of this edit. Executing 08 needs
`news_features.parquet`, which this worktree did not have, so 07 was run to build it - and 07
does not reproduce. Two consecutive runs in one environment disagree on five columns. The
commit records every moved value old -> new with the cause. Per Stefan's ruling the notebooks
demonstrate process, data and models change, and a changed number is not a decision, so the
numbers land where they land. Every claim 08's prose makes was checked against the new render
and still holds; the prose quotes no specific value, so nothing had to follow it.

nbcheck reports 0 findings on all four renders.


stefan-jansen:

PR #979, opened 2026-09-11
Check every supersedes declaration before the fits, against the registry the run will write

stefan-jansen:
A `SUPERSEDES_*` literal that no longer names a live generation is refused at the
freeze - after the fits are paid for - and is invisible until then. An unchanged
re-run matches on members and never reads the declaration, so the literal sits
dead through any number of green runs and fails the first one that moves the
set's membership, which is the refit the notebook exists for. On 2026-09-06
twelve `sp500_options` notebooks passed smoke and three failed the real chain
fifteen minutes later on eleven stale literals.

`scripts/check_supersedes_literals.py` has been able to answer this since #944
and nothing called it before a run. This wires it into `scripts/pre_run_gate.py`,
scoped to the notebook being queued when one is named: a dead literal in a
sibling notebook is somebody else's run to fix and must not block this one.

It cannot live in CI. The check needs `run_log/`, which is gitignored, so a
checkout has no registry to ask. The gate has one.

## Reading the registry the run will actually write

The first commit asked the checker for its own default artifacts root,
`~/ml4t/artifacts/case_studies`. That is where a standard checkout's `run_log/`
symlink happens to point and nowhere else. `check_registry_ready`, ten checks
earlier in the same gate, resolves the registry through `get_case_study_dir`,
which honours `ML4T_OUTPUT_DIR` and a relocated `CASE_STUDIES_DIR`. Under any
redirection the two checks in one gate read two different registries.

The expensive half is the hardcoded location holding nothing. `check_case_study`
records a missing registry as a `no-registry` finding, `no-registry` is not
`stale`, so the gate reported `no dead declaration` for a check that never opened
a database. That is the absence-reading-as-a-pass this check was added to remove,
reproduced inside the check itself. Both halves are fixed in `86ceb534`: the root
comes from `get_case_study_dir(case_study, create=False).parent`, and an absent
registry is a failure naming the path it looked at.

`tests/test_pre_run_gate_supersedes.py` builds a three-generation chain reachable
only through `ML4T_OUTPUT_DIR` and declares the generation two back, which the
head no longer offers. Both defect tests were verified failing before the fix
with the exact symptom, `[PASS] ... no dead declaration`. The live-literal control
passes before and after, which is why it is kept: before the fix it passed
vacuously.

Against the real registries the fixed gate reports `fx_pairs/14_portfolio_management`'s
three dead declarations, passes `etfs`, and fails with the path when there is no
registry.

## Two things this deliberately does not do

**It does not fix the dead literals it finds.** There are eight across three case
studies. Correcting one makes the paired `.ipynb` stale, and the sanctioned way
to commit that re-executes the notebook. A canary render
(`cme_futures/06_linear`, exit 0) left the shared registry byte-identical across
all four tables, so a warm re-run is safe - but its one output change was
`280 folds fitted, 0 reused` becoming `0 folds fitted, 280 reused`, which is the
exact symptom of the open wording issue about reuse counts. Re-rendering now buys
a second re-render later to correct the wording. The gate makes deferring safe:
these declarations are now refused at queue time rather than at the freeze.

**It does not gate on `undeclared`.** A live generation that no declaration names
at all is a larger and separately filed problem; this check reports only `stale`.

## Also in this branch

The comment above `SUPERSEDES_CANDIDATE_SETS` in
`case_studies/fx_pairs/14_portfolio_management.py` described the literals as
current. Landed through `sync-prose`, so no re-render.

`test_the_scan_reads_the_real_corpus_without_falling_over` has been red locally
since `_undeclared_heads` landed, because `undeclared` was missing from its status
allowlist. The test exists to catch a checker that invents a status; `undeclared`
is one the checker declares and `main()` reports and exits on, so the allowlist
was stale rather than the checker wrong. The test skips in CI, where `run_log/`
does not exist, so the failure was local only.


stefan-jansen:
Superseded by #983. Same two commits cherry-picked onto a fresh branch off current main, patches verified byte-identical; the only tree difference is that #983 sits on b390e8c3 rather than e75dcac9.

Moved because this branch's name carries an agent-workspace issue number, which shows on the PR and stays in its record after the branch is deleted, and a name cannot be sanitized at merge the way a commit message can. The two commit subjects carried the same bare number and are reworded on the new branch.

PR #982, opened 2026-09-11
ch02: run the two data-management notebooks against a deterministic provider in CI

stefan-jansen:
`02_financial_data_universe/18_data_management` and `19_incremental_updates` are
the only chapter-2 notebooks that still fetch from a live vendor on every
commit. `16_provider_comparison` and `17_complete_pipeline` were moved to the
weekly tier for exactly this reason; these two were not.

That has made `ch02-03` red three times on work that touched neither notebook:
twice on a null `open` in Yahoo's current-session placeholder bar (2026-08-29
and 2026-09-04, both on `main`), and once today on a rate limit that the
provider reported as an invalid symbol, caught on a PR whose only change was to
`case_studies/utils/backtest_explorer.py`. None of those reds were about the
storage, validation or incremental-update code these notebooks teach.

## What changed

Both notebooks already declare `DEMO_PROVIDER` in their parameters cell. 18 was
calling `provider="yahoo"` directly at three of its six fetch sites, so an
override could only ever have reached half the notebook; those three now read
the declared parameter. `tests/overrides.yaml` sets `DEMO_PROVIDER: mock` for
both, which is ml4t-data's deterministic provider.

The reader's default is unchanged. `DEMO_PROVIDER` is still `"yahoo"` in both
notebooks, the committed outputs are still a real Yahoo run, and the prose about
the placeholder bar and the retreat past it still describes what the committed
output shows.

## Verification

- Both notebooks pass under the override: `2 passed in 90.21s`, against `382.72s`
  for the live path.
- Both still pass with `HTTP_PROXY` and `HTTPS_PROXY` pointed at a dead port:
  `2 passed in 90.65s`. This is what establishes that the override reaches the
  provider, rather than Yahoo merely being healthy at the time.
- `tests/test_pm_helpers.py` and `tests/test_overrides_schema.py`: 126 passed.
  The first includes the gate that a declared parameter actually reaches its
  notebook.
- 18 was re-executed for its provenance stamp. The only output change is
  924 -> 925 rows per symbol, one trading day since the previous run.


stefan-jansen:
Second commit is unrelated to chapter 2 and corrects a comment I put in this
file yesterday.

It claimed `test_one_producer_per_fixture_file` and
`test_every_fixture_file_has_a_producer` were named by no job and had both been
passing vacuously. Wrong about the mechanism: `test-unit`'s last step has swept
the rest of `tests/` by subtraction since 2026-08-24, ignoring only what
`.github/ci/unit-test-quarantine.txt` lists, so an unnamed file does run.

Measured by reproducing that step's environment - `ML4T_DATA_PATH` on the repo's
own `data/`, no test-data checkout:

```
tests/test_one_producer_per_fixture_file.py .......                      [ 36%]
tests/test_every_fixture_file_has_a_producer.py ssss                     [ 57%]
tests/test_fixture_builders_require_their_inputs.py .........
SKIPPED [4] not a test-data checkout (no manifest.json)
```

So `test_one_producer_per_fixture_file` was already passing, and
`test_every_fixture_file_has_a_producer` was a real gap for a different reason
than the one I gave: it collected and skipped all four of its tests, and
`test-unit` does not fail on a skip. Moving it to `test-unit-data` was the right
fix; the stated cause was not. The older sentence above it - "Every job in this
file selects by explicit path, so a file nobody lists never runs" - predates the
sweep and is corrected in the same hunk.

PR #986, opened 2026-09-11
A worktree path is a machine path, and the sanitizer was writing one it could not then see

stefan-jansen:
The sanitizer maps each repo root to `""` so paths come out repo-relative. A worktree is a
repo root too, and there was no rule for one, so
`/home/stefan/ml4t/public-teach-e/10_text/out.parquet` fell through to the generic
`/home/[^/]+/ml4t/` rule and came out as `~/ml4t/public-teach-e/10_text/out.parquet`. The
username is gone, the lane name is not.

**Nothing downstream could see the result.** `nbcheck`'s `ABSOLUTE_PATH` and this repo's own
`guards` check both match the `/home/` spelling. The leak was detectable right up until the
tool whose job is to remove it ran, and undetectable afterwards, which is why it accumulated
silently. A reader opening `case_studies/us_equities_panel/04_model_based_features.ipynb`
saw a directory that exists on one machine and belongs to a lane rather than to the book.

## Scale, measured at b390e8c3 across 483 tracked notebooks

| where | files | occurrences |
|---|---|---|
| `metadata.papermill.input_path` + `output_path` | 362 | 724 |
| cell outputs | 48 | 374 |
| cell source | 0 | 0 |
| absolute `/home/<user>` anywhere | 0 | 0 |

The last row is why a sweep for the absolute form reports the corpus clean: the `guards`
check has always enforced it. 27 distinct worktree spellings, led by
`public-s6-us_equities_panel` (231) and the bare main checkout `~/ml4t/public` (166).
`04_model_based_features` holds 219 of the 374 reader-visible occurrences by itself.

## Two rules, and the negative cases written first

The `/home/` form covers a fresh render. The `~/` form covers the 362 files the generic rule
has already rewritten - **without it the sweep finds nothing to fix**. `(?=/)` after the
optional lane suffix is what keeps a hypothetical `~/ml4t/publications/` out; a bare
`public[^/]*` eats it.

Six cases that must fire, five that must not: `publications`, `~/ml4t/case_studies/...`, a
bare `~/ml4t/`, `~/ml4t/public` with no trailing slash, and the word "public" in ordinary
prose.

## `sync-paths`, because a re-run is the wrong operation

A path in a cell output is hashed by `outputs_digest`, so the sanitize reads as stale on
those 48. No existing tier covers it: `sync-prose` refuses because the outputs moved,
`sync-alt` writes alt text, a figure title is a code change.

Re-running is not merely expensive here. **Re-executing cannot remove a worktree path from
an output - it writes the current worktree's path instead** - so the gate would be demanding
the one action that reintroduces the defect. That is the argument `VOLATILE_OUTPUT_KEYS`
already makes for figure `alt`: re-executing cannot change the image either.

`notebook_provenance.py sync-paths` re-stamps `outputs_digest` and keeps `source_py_blob`,
`executed_at` and `executor`, because the `.py` did not change and the outputs are still the
ones that run produced. It is decidable rather than trusted: it reads the committed
notebook, applies the sanitizer, and requires the result to equal the file on disk exactly.
A number that moved, a cell that was edited or a figure that was regenerated fails that
comparison and is refused.

## Verification

All 362 rewritten notebooks have cells exactly equal to the committed file with the
sanitizer applied, checked by re-deriving each one rather than by reading the diff. 314 are
metadata-only and their digests do not move; 48 moved and were re-stamped. The sanitizer
reports the tree clean and `test_no_machine_specific_paths_in_committed_notebooks` passes.
No `.py` is touched - `source` is outside what the sanitizer may rewrite.

Ten notebooks carry a stamp with no `outputs_digest` key at all - nine
`01_feasibility_analysis` plus `nasdaq100_microstructure/02_labels` and `/03`. They were
stamped before that field existed, and `notebook_provenance.py check` classifies them as
"stamped before the digests existed" rather than stale. They need no re-run and this sweep
does not change them. Noted because a verification script that compares a stamp's digest to
the recomputed one reads a missing key as a mismatch, which is how I first counted them.

## Merging this conflicts with every open PR that touches a notebook

362 rewritten notebooks is a wide diff by construction, and it cannot be split: the required
`guards` check asserts the tree is already clean of whatever the sanitizer would rewrite, so
the rule and the sweep are one commit. Anything merged after it that carries a notebook
rendered before it will need the sanitizer re-run on that file.


stefan-jansen:
Correction to the commit message on 4e9388cc, which the PR body now carries in fixed form.

The commit says ten notebooks "carry a stamp that already disagreed with their own outputs at HEAD". That is not what they are. All ten stamps have no `outputs_digest` key at all - they were written before that field existed - and `notebook_provenance.py check` classifies them as "stamped before the digests existed", a known advisory category. They are not stale and need no re-run.

My verification script compared `stamp.get("outputs_digest")` to the recomputed digest, so a missing key read as a mismatch. The finding that matters is unaffected: zero notebooks were made stale by this sweep, and all 362 rewrites are exactly the committed file with the sanitizer applied.

Since the squash body is composed from commit messages, this one should be merged with `gh pr merge --squash --body` carrying the corrected paragraph rather than the commit's.

PR #987, opened 2026-09-11
deps: move ml4t-data from 0.1.2 to 0.1.6

stefan-jansen:
`uv.lock` pinned `ml4t-data` at 0.1.2. `uv sync --locked` is what CI and the
install instructions both use, so four releases of the library every chapter
imports reached neither a CI run nor a reader. `pyproject.toml` already said
`>=0.1.2`, so the constraint was never what held it back - the lock was.

Measured in a fresh worktree built from the repo's own lock:

```
$ uv run python -c "import importlib.metadata as m; print(m.version('ml4t-data'))"
0.1.2
```

## What the four releases carry

Two are chapter-2 defects a reader hits:

- **0.1.4** drops the current session's placeholder bar instead of raising on it.
  A window containing that row alongside real bars now succeeds, which is the
  behaviour `18_data_management` and `19_incremental_updates` document in prose.
- **0.1.6** stops reading an empty `yf.download` frame as "symbol not found".
  Yahoo returns that one shape for a rate limit, a network error and a delisted
  ticker alike, so a throttled reader was told `AAPL` does not exist. The empty
  branch now refetches through `Ticker.history(raise_errors=True)` and reports
  what the vendor actually said.

0.1.3 is an ETF partition rebuild and a set of CI chores; 0.1.5 names the day and
reason in an async Binance download failure.

## Scope

Only `ml4t-data` moves. The lock diff is its four lines - specifier, version,
sdist, wheel - and nothing else resolved differently.

`ml4t-diagnostic`, `ml4t-engineer`, `ml4t-backtest` and `ml4t-live` are also
behind and are deliberately not here. Each needs its own reading of what moved,
and two of them carry behaviour changes that want a measurement before they land.

## Verification

- `uv sync --locked --extra dev` resolves 0.1.6 with no other change.
- `tests/test_env_matches_lock.py` and `tests/test_toolchain_pins.py`: 23 passed.
- Both chapter-2 notebooks that fetch live pass against it: `2 passed in 94.25s`.
  This branch does not carry #982's deterministic-provider override, so that run
  went to Yahoo for real, which is the point of running it here.

Read the `ch*` and `cs-*` jobs on this one rather than the five required checks:
a library behaviour change surfaces in a notebook, not in `test-unit`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -119,6 +119,17 @@ jobs:
               - 'case_studies/config/**'
               - 'data/**'
               - 'pyproject.toml'
+              # And the lock, which is what `uv sync --locked` actually installs.
+              # A `uv lock --upgrade-package <x>` that does not touch
+              # pyproject.toml changes the library version every notebook imports
+              # and, without this line, matches only `image-inputs` - so the whole
+              # chapter and case-study matrix is skipped and the PR reports green
+              # on `test-unit`, which pins nothing about notebook behaviour.
+              # Dependabot opens no PRs against this repo
+              # (.github/dependabot.yml, open-pull-requests-limit: 0), so this
+              # costs a full matrix only on a deliberate bump, which is the case
+              # that needs one.
+              - 'uv.lock'
               - '.github/workflows/test.yml'
               - 'matplotlibrc'
               - 'tests/__init__.py'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -53,10 +53,13 @@ dependencies = [
     # ===================
     # ML4T Libraries (PyPI)
     # ===================
-    # ml4t-data 0.1.2: HiveStorage.partitions() reports what a write committed, the anomaly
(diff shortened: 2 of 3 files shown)


stefan-jansen:
Second commit is the gap this PR walked into.

`shared` in `test.yml` lists `pyproject.toml` but not `uv.lock`, and
`uv sync --locked` is what installs. So `uv lock --upgrade-package <x>` with no
`pyproject.toml` edit moves the library version every notebook imports while
matching only the `image-inputs` filter: `all` stays false, the whole chapter and
case-study matrix is skipped, and the PR reports green on `test-unit`, which pins
nothing about notebook behaviour.

This PR happens to edit `pyproject.toml` as well, which is the only reason its
own matrix fans out. A bump that touched the lock alone - the ordinary shape of
a dependency refresh - would have merged on a run that executed no notebook
against the new library.

Dependabot opens no PRs against this repo: `.github/dependabot.yml` sets
`open-pull-requests-limit: 0` and repository-level security PRs are off. So the
added line costs a full matrix only on a deliberate lock change.

PR #988, opened 2026-09-11
Every reuse disclosure goes through one helper, so none opens on a zero

stefan-jansen:
On a warm registry a sweep summary read:

```
fwd_ret_8h / ew_top3: 262 backtests registered
  this execution: 0 computed, 262 served from the registry
```

To someone reading the chapter rather than the run, that first clause says the backtest section
computed nothing. What happened is that an earlier execution computed all 262 and this one reused
them, which is the intended idempotent behaviour. The disclosure has to stay - a notebook that
serves results from the registry without saying so is a silent skip - so this is wording.

`research.reuse_disclosure` already held the corrected wording and was adopted at six sites. This
finishes it: eight more notebooks stop spelling their own version.

```
reused all 262 from the registry, computed by an earlier run
computed all 262
computed 40, reused 222 from the registry, 3 failed
```

Fourteen sites total - four through `execution.disclosure()`, ten through the free function over
counts the notebook tracks itself.

## The three that were a count list rather than a pair

Two of them opened on a zero the same way:

| notebook | warm output before |
|---|---|
| `nasdaq100_microstructure/14_backtest` | `Sweep complete: completed=0, skipped=500, failed=0 in 12s` |
| `sp500_equity_option_analytics/14_backtest` | `Sweep complete: 0 run in 12s (0 failed, 500 already complete)` |

`us_firm_characteristics/11_backtest` led with its total so it read correctly, but it called reuse
`skipped` while every other case study called it something else. All three route through the helper
now, with the counts read off each loop's own variables - `results` there holds only what ran,
because a skipped identity `continue`s before the append.

## The guard, and what it deliberately does not ban

Fixing one notebook and leaving the rest is worse for a reader than the old phrasing, because that
notebook becomes the one case study that words it differently. So the regression worth catching is
the *next* sweep spelling its own pair, which no unit test of `reuse_disclosure` can see. The guard
bans a count interpolated directly before `computed` or `served from the registry`, and a negative
selftest runs the exact line this was filed against through the same expression - so a scan that
finds nothing is distinguishable from a pattern that has stopped matching.

It does not ban the model notebooks' `{n} configurations: {fitted} folds fitted, {reused} reused`.
That leading count is the configuration total, which is never zero, so warm it reads correctly:
twenty configurations, none refitted, five hundred folds reused. Same for the `Planned {n}
backtests; {m} already complete` lines, which state a plan before a sweep rather than what the sweep
did. Neither has the defect, and banning them would have cost twenty-one model notebooks a render
for nothing.

## Renders

Five of the eight notebooks carry no provenance stamp, so the change syncs into them and claims
nothing new. `fx_pairs/13_backtest`, `fx_pairs/14_portfolio_management` and
`us_firm_characteristics/14_costs` went STALE on a printed string that cannot move a computed value;
they are cleared rather than re-run, because all three sit downstream of a baseline read whose
population scoping is still being decided, and a render spent now would be spent twice.

`notebook_provenance.py check --since origin/main` reports 0 stale, 0 contradicted, 7 unverified
(advisory). `tests/test_research_contract_execution.py` passes, 42 tests.

https://claude.ai/code/session_01EKVuCYKyzMyRG6SEm8HwqU


stefan-jansen:

PR #989, opened 2026-09-11
Every file in the CI fixture has a producer, and three of the last six were wrong

stefan-jansen:
The fixture backlog - files under no `Dataset.owns` and in no manifest entry - closes here. It was 149 of 327 when it was first counted, 27 after the factor/macro/on-chain tranche, and this branch takes the last of it: 21 deletions at paths nothing resolves, five declarations, and one more deletion that only showed itself when its builder was run.

**A fixture with no builder has never been compared to production, and three of the last six turned out to be wrong.** That is the argument for declaring rather than recording. None of the three announced itself, because each rendered an empty frame rather than raising.

- `equities/market/sp500/options_eda/year={2019,2020}.parquet` carried moneyness 0.95-1.05 and nothing past 45 days to maturity, against `02_financial_data_universe/07_sp500_options_eda`'s own declared `CHAIN_BAND` (0.7, 1.3), `SPREAD_BAND` (0.5, 1.5), `WING_BAND` (0.9, 1.1) and `SURFACE_MAX_DAYS` 180. So the smile had no smile, the wing analysis had no wings, "away from the money" was empty and the far-expiration section drew nothing. Its symbol list was wrong the same silent way - AMD, GOOG and INTC where the loader's docstring and `08_financial_features/03_structural_cross_instrument_features` line 73 name BA, JPM, KO and XOM, so four of the eight symbols that notebook asks for by name came back empty.
- `alternative/text/financial_phrasebank/sentences_allagree.parquet` held 4,846 rows - the whole corpus at 50% annotator agreement - under the filename that promises the 2,264 unanimous ones. `10_text_feature_engineering/04_bert_finetuning` fine-tunes on this file, so CI trained on 2,582 sentences the annotators disagreed about, each carrying whatever label the corpus assigned. `load_financial_phrasebank(agreement=...)` looks like it would have caught that and does not: it filters an `agreement` column neither production's file nor the fixture's has, so the argument has never selected anything.
- `prediction_markets/polymarket_events.parquet` held three events against production's 22, which is fewer rows than the panel `13_polymarket_prediction_markets` groups by category.

`institutional_holdings_13f_bulk` was correct already, and its declaration is worth having for a different reason: filtering production to the 600 CIKs with the largest reported value reproduces the fixture row for row - 1,327,852 rows, 743 filings - and that rule had never been written down.

**`nasdaq100_taq` is the one the builder caught.** It was declared as redistributed AlgoSeek tick data with no source to reduce from, and the first run of its own contract refused it: five symbols where the contract asserted one. Measured, it is 20,000 rows, five symbols at exactly 4,000 each over twenty March 2020 sessions, every consecutive gap exactly 117 seconds, `event_type` "trade" on every row and no quote events at all. `load_nasdaq100_taq` reads `equities/market/microstructure/trade_and_quotes/symbol=*`, which `generate_test_microstructure.py` writes and which carries 600 rows on the one session the notebooks name, with TRADE, QUOTE BID and QUOTE ASK events - which is what `12_algoseek_taq_lob_reconstruction`'s Lee-Ready classification needs and what this file has none of. Nothing in the repository resolves `nasdaq100_taq/`. So it is a deletion, and `Dataset.irreplaceable` goes with it: it existed for this one dataset and now has no user.

It leaves the fixture in the next change rather than this one, because removing it from `UNPRODUCED` and deleting it from the test-data repo cannot land together - whichever half arrives first turns a test red on every branch that has not merged the other.

## What this costs

`options_eda` goes 25 MB -> 118 MB, on a 587 MB fixture. Both axes that would cut it are pinned: `02/09` builds a daily straddle series and needs every 2019 session, and `08/03` names all eight underlyings.

**Measured, because the row count creates the wrong intuition: the growth is almost entirely rows these notebooks never plot.** `02/07`'s every plotting path is bounded by one symbol on one session, so on the real `sample_date` (`options["timestamp"].max()`, 2020-12-31) it sees 680 rows in `aapl_day` and 217 in `aapl_calls`, and the `trendline="lowess"` scatter - the one superlinear element in the notebook - is handed **34 points**. The two whole-panel aggregations take 0.21 s and 0.24 s. The entire cost of the forty-fold panel is the load, 0.54 s against about 0.1 s before.

**And the same number states the original defect in cost terms.** Under the old 0.95-1.05 band that lowess call got **6 points** at the earliest session. A trendline through 6 points is not a smile. So the old fixture was not buying speed with its narrowness - it was drawing an empty analysis for free, which is why nothing ever flagged it.

## Pairs with a test-data push

`ml4t/third-edition-test-data` `6dbc3fa0` carries the rebuilt files and the reconciled manifest. It had to go first: `test_the_manifest_subsets_are_exactly_the_declared_ones` compares that manifest against each branch's own `DATASETS`, so until this merges the five new subsets show as undeclared on `main` and on every open branch. The instruction in that window is `git merge origin/main`, never a re-run - a re-run reads the same pre-merge file and fails identically.

## Renders this invalidates

Five notebooks now draw against different data, and none is in any pane's current scope: `02/07_sp500_options_eda`, `02/08_options_greeks_computation`, `02/09_options_continuous`, `08/03_structural_cross_instrument_features` and `10/04_bert_finetuning`. Their committed renders were produced against the fixtures corrected here, and the sections at issue drew empty frames rather than failing - so a reader sees nothing wrong and CI stayed green. Their register rows reading `done` is a dated claim, not a fact, and they need re-opening deliberately.

https://claude.ai/code/session_01NReuqDy57ft7qkAAKUEtVG

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1399,14 +1399,12 @@ jobs:
             cp test-data/intermediates/ch20_synthesis/output/* 20_strategy_synthesis/output/
             echo "Seeded Ch20 synthesis outputs"
           fi
-          # Ch24/11 research_operator: artifacts dir is gitignored in this repo;
-          # source them from test-data into the notebook dir.
-          if [ -d test-data/data/autonomous_agents/operator_artifacts ]; then
-            mkdir -p 24_autonomous_agents/operator_artifacts
-            cp test-data/data/autonomous_agents/operator_artifacts/*.json \
-              24_autonomous_agents/operator_artifacts/
-            echo "Seeded ch24/11 operator artifacts"
-          fi
+          # Ch24/11's two operator traces are not seeded here. They are tracked in
+          # this repo under 24_autonomous_agents/operator_artifacts/ (force-added
+          # past the .gitignore), so the checkout already carries them and CI reads
+          # what a reader reads. The copy that used to stand here overwrote both
+          # with the test-data fixture's unsanitized originals, which carry the
+          # author's absolute home directory in every skill path.
 
       - name: Run chapter notebooks (${{ matrix.name }})
         if: steps.reach.outputs.have_key
--- a/data/equities/firm_characteristics/dataset_card.ipynb
+++ b/data/equities/firm_characteristics/dataset_card.ipynb
@@ -976,8 +976,8 @@
    "end_time": "2026-09-11T18:55:00.052207+00:00",
    "environment_variables": {},
    "exception": null,
-   "input_path": "~/ml4t/public-fixture-d/data/equities/firm_characteristics/.dataset_card.build.1734332.ipynb",
-   "output_path": "~/ml4t/public-fixture-d/data/equities/firm_characteristics/.dataset_card.papermill.1734332.ipynb",
+   "input_path": "data/equities/firm_characteristics/.dataset_card.build.1734332.ipynb",
+   "output_path": "data/equities/firm_characteristics/.dataset_card.papermill.1734332.ipynb",
    "parameters": {},
    "start_time": "2026-09-11T18:54:50.834110+00:00",
    "version": "2.7.0"
--- a/data/equities/firm_characteristics/dataset_card.ipynb
+++ b/data/equities/firm_characteristics/dataset_card.ipynb
@@ -2,13 +2,13 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "id": "d381d65f",
+   "id": "e60444b1",
    "metadata": {
     "papermill": {
-     "duration": 0.006943,
-     "end_time": "2026-05-03T01:59:03.910766+00:00",
+     "duration": 0.003201,
+     "end_time": "2026-09-11T18:54:51.810924+00:00",
      "exception": false,
-     "start_time": "2026-05-03T01:59:03.903823+00:00",
+     "start_time": "2026-09-11T18:54:51.807723+00:00",
      "status": "completed"
     }
    },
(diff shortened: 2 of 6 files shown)


stefan-jansen:
Correction to the "Renders this invalidates" section above: **it invalidates none of them.** I asserted those five renders were produced against the fixtures corrected here. They were not - they were produced against real production data, and they are correct as committed.

Checked from the committed `.ipynb` outputs rather than from configuration, since the data root is per-worktree:

| notebook | production symbols in outputs | fixture-only symbols (AMD/GOOG/INTC) |
|---|---|---|
| `02/07_sp500_options_eda` | 57 hits across all eight | 0 |
| `02/08_options_greeks_computation` | 10 | 0 |
| `02/09_options_continuous` | 1 | 0 |
| `08/03_structural_cross_instrument_features` | 43 | 0 |

The pre-#989 fixture carried AMD, GOOG and INTC and could not have produced a render naming BA, JPM, KO and XOM. `10/04_bert_finetuning`'s render reports **2,264** sentences, which is production's unanimous corpus - the old fixture held 4,846, so that render cannot have come from it either. All five stamps also read `executor: local-uv`, and all five predate the test-data rebuild (`6dbc3fa0`, merged 21:11Z) by hours.

So what this PR fixed is what **CI** exercises, not what a reader sees. That is still the whole point of the change - three of the last six fixture files were wrong and each drew an empty frame rather than failing - but the register rows for those five stay `done`, and nothing needs re-running.

The one thing that did survive as real work is the loader defect noted in the body: `load_financial_phrasebank(agreement=...)` filters an `agreement` column no distributed file carries, so the argument has never selected anything. Fixed in #996.

PR #991, opened 2026-09-11
A supersedes declaration names the lineage, not a generation of it

stefan-jansen:
A `SUPERSEDES_*` literal is a committed constant that has to equal a value the registry moves on every publish. Measured over the corpus today: 48 declared literals, of which **6** name a value `create` accepts, **19** name the generation it has already replaced, **8** name a generation no lineage holds, and **5** cannot be placed at all.

## The 19, which nothing reported

`population_supersedes` offers a declared hash when it equals `current.supersedes`, and `scripts/check_supersedes_literals.py` read that as `live: names what the tip replaced; a re-run resolves to it`. That is true of an unchanged re-run and false of every other run. `create` accepts the tip and nothing else, so the run whose membership **moves** is refused - and that is the refit a chain is queued for.

`tests/test_supersedes_declares_intent.py` is the reproduction, end to end through the real `create` on both lineages:

```
gen1: 62ece7eb073f
gen2: bfa25aff19e1 supersedes 62ece7eb073f
candidate_set_supersedes(declared=gen1) -> 62ece7eb073f      # the resolver offers it
CandidateSet.create(members=[m1, m2, m3], supersedes=62ece7eb073f)
  -> ValueError: a changed candidate set named 'the-set' must explicitly supersedes bfa25aff19e1
```

Identical on the population path. Those declarations are now reported `behind` and refuse exactly as `stale` does: 8 refusals across the corpus became 27.

The two refusals do not cost the same, and the difference is measured rather than assumed. `run_official_models` writes the population snapshot before the first fit, so a population refusal lands during planning; a candidate set is frozen in the last cell, after every fold is fitted. Split of the 27: **5 behind + 4 stale on the candidate-set lineage**, refused after the fit; **14 behind + 4 stale on the population lineage**, refused during planning.

#983 wired the checker into `scripts/pre_run_gate.py` and keyed it on `is_stale`, which is `status == "stale"` exactly - so the gate ran, saw 19 real refusals, and reported `no dead declaration`. Fixed here, with a test on the three-generation chain that PR's own fixture already builds.

## `SUPERSEDES_LIVE`

Both resolvers are already called with `name=`. The predecessor is therefore a lookup under a name the call site holds, not something the source has to quote, and the record of which generation was replaced is written by `create` into `supersedes_hash` - where a reader looks for a lineage anyway. It was never recoverable from source.

A declaration of `"live"` resolves to the tip at run time and is reported `intent`. It is additive: every existing hash resolves exactly as before, so no notebook has to change to stay correct. What changes is the repair the checker prints - `"live"` rather than the current head, since pasting a head is correct only until the next publish of whatever freezes that lineage.

It weakens no refusal, and each of these is a test:

- a lineage this run does not name is still refused when its membership moves;
- a hash two or more generations back is still withheld;
- a hash in no lineage is still withheld;
- a clean clone is handed nothing, so a reader still publishes generation one;
- a narrowed run under its own name is handed nothing;
- passing the sentinel to `create` past the resolver is still refused as a first version.

Verified as a negative control: with the two resolver branches commented out, three tests fail and the other eight pass, which is what makes them selftests of the sentinel rather than of the surrounding code.

## Also

The corpus scan's status allowlist is now asked of the script instead of restated in the test. That second copy went stale once already - `undeclared` reached it only in #983, long after the checker declared it - and it can only go stale unnoticed, because the test skips wherever `run_log/` is absent, which is every CI run.

## Not in this PR

Nine declarations in `fx_pairs` and `sp500_equity_option_analytics` that no checker can classify - the hash is in no lineage AND the name is built at run time, so nothing can say whether the literal is dead or waiting for a first publication. They are swept and verified against the live registries (the corpus then holds zero unresolvable), but a parameters-cell value is a code cell, so five paired `.ipynb` files go stale and the provenance gate refuses the commit. Three of the five last exited 1, so a re-run restores nothing until those failures are diagnosed. Held as a patch for the case-study owners, who own the re-render-versus-clear call.

https://claude.ai/code/session_01GwvExyYgecwyTLrW2F4Juw

### DIFF
--- a/case_studies/research/__init__.py
+++ b/case_studies/research/__init__.py
@@ -48,6 +48,7 @@ from .model_planning import (
 )
 from .models import ModelRequest, ModelRun, ResolvedModelRequest
 from .population import (
+    SUPERSEDES_LIVE,
     OfficialPopulation,
     RetirementSplit,
     population_supersedes,
@@ -160,6 +161,7 @@ __all__ = [
     "unfinished_sweep_plans",
     "planned_backtests",
     "upstream_plan_hashes",
+    "SUPERSEDES_LIVE",
     "population_supersedes",
     "research_name",
     "resolved_model_plan",
--- a/case_studies/research/comparison.py
+++ b/case_studies/research/comparison.py
@@ -13,7 +13,7 @@ import polars as pl
 from case_studies.utils.registry.specs import canonical_json, compute_hash
 from case_studies.utils.registry.store import _git_hash, _open_registry, _utc_now
 
-from .population import _refuse_preview_activation
+from .population import SUPERSEDES_LIVE, _refuse_preview_activation
 from .results import Result
 
 if TYPE_CHECKING:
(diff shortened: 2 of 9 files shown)


stefan-jansen:

PR #992, opened 2026-09-11
ch05/06: an empty GReaT sample is reported as an empty sample, not as a missing column

stefan-jansen:
`05_synthetic_data/06_llm_tabular_great` failed CI with

```
Error: Cell 21 (KeyError): "None of ['feature'] are in the columns"
```

which names the column the cell was about to write rather than the condition that produced it. `numerical_rows` had come out empty, so `pd.DataFrame([])` had no columns at all and `.set_index("feature")` raised.

## Reproduced first, three ways

Driving the shipped cell with an empty frame, with the schema at zero rows, and with rows whose numerics are prose gives that exact `KeyError` in all three.

Running the notebook at the CI parameterization (`EPOCHS=1, N_SAMPLES=50, N_GENERATE=10`) returns ten rows reading `"the most common mod"`, `"not a good idea"`, `"not the case"` and prints `Insufficient valid synthetic samples (0)` - one parseable column short of the failure itself. An empty result is a foreseeable outcome of `great.sample`: an undertrained backbone writes text where a number belongs, `pd.to_numeric(errors="coerce")` turns it into NaN, and every numeric column can lose every value at once.

## The parse check was measuring the wrong thing

The cell headed *"Check for parsing errors (NaN values)"* counted nulls. A cell holding the words `"not the case"` is a string the row parser left in place, not a null, so the check reported clean on exactly the failure it is named for. It now counts, per column, how many values convert to a number, and names any schema column the sampler did not return. Both counts are taken once and the later sections read them.

## What changed

| cell | before | after |
|---|---|---|
| parse check | counted nulls | counts unparseable values, names absent columns |
| numerical comparison | `set_index("feature")` on a possibly empty frame | builds the table when it has rows; otherwise prints which columns hold no parseable value and which are absent |
| categorical comparison | rendered a blank table silently | says so when empty |
| TSTR split | `synthetic_df[feature_cols]` | `.reindex(columns=...)`, so an absent column arrives all-null and reaches the existing guarded branch instead of raising a second missing-column error 130 lines later |

## Flake, not regression

The issue left that open. The CI evidence settles it: the notebook blob is byte-identical across the five `ch05` jobs run today, and `tests/overrides.yaml` is identical between the failing sha and a passing one. Same inputs, one failure, four passes.

## Verification

The four changed cells were driven against all three empty shapes plus a control of usable rows. None raises, and the control still yields the 6x4 comparison table and 40 synthetic training rows.

Executed at production parameters - no reduction, `parameters: {}` in the provenance stamp - and the normal path is unchanged: 500 of 500 rows parse in all six columns, no guard fires, TSTR ratio 99.8%. 27m31s, 1.7 GB peak, through `nb-run.sh`.

`guards` locally: pair sync, import resolution (1,130 files), constant subscripts (selftest 26/26), notebook output hygiene (21 passed), and `notebook_provenance.py check` reporting `0 stale, 0 contradicted, 0 outputs-changed`.

## Expected reds not from this branch

- `test-unit-image`: #987 moved ml4t-data 0.1.2 to 0.1.6 on main while the published image still had 0.1.2. Cleared by a re-run once the image carries 0.1.6.
- `test_the_manifest_subsets_are_exactly_the_declared_ones` and `test_no_exempt_path_has_left_the_fixture`: cleared by merging `main` after #989 lands, never by a re-run.

https://claude.ai/code/session_019xcyuaQpWc4fHFL45FjARQ


stefan-jansen:

PR #1005, opened 2026-09-12
CI: cache HuggingFace checkpoints for the chapter jobs

stefan-jansen:
Six chapter jobs download model checkpoints from HuggingFace while the notebook runs, and nothing cached them. Every job of every run fetched them again over an unauthenticated connection, several jobs at once from the same runner pool, and HuggingFace answers that with a 429. The failure then lands on whatever commit happens to be under test.

## Measured

| when | job | branch | error |
|---|---|---|---|
| 2026-09-11 22:57Z | `ch11-12` | #994 | `429 Client Error: Too Many Requests` loading `yiyanghkust/finbert-tone` |
| 2026-09-12 01:12Z | `ch13` | `main` | `We couldn't connect to 'https://huggingface.co'` loading `ibm-granite/granite-timeseries-ttm-v1` |

Neither PR touched a notebook in the chapter that failed. #994 is a dataset-card change under `data/`; the `main` run was a lockfile merge.

## What downloads what, measured from the first full run

| job | cached | |
|---|---|---|
| `ch11-12` | 777 MB | `yiyanghkust/finbert-tone` |
| `ch05` | 291 MB | `distilgpt2` |
| `ch13` | 197 MB | `ibm-granite/granite-timeseries-ttm-v1`, `amazon/chronos-t5-tiny` |
| `ch10` | the four text models and two datasets | |
| every other `ch*` | ~200 bytes | downloads nothing in CI |

Four jobs of eighteen, and they are the ones that were failing. An earlier version of this table was grepped from `"org/name"` literals in the notebooks rather than measured, and got two things wrong: it missed `ch05`, whose `distilgpt2` has no org prefix, and it listed `ch04`, `ch22` and `ch23`, which name `all-MiniLM-L6-v2` in source but whose notebooks are deselected or skipped in CI.

No case study touches HuggingFace, so this is the chapter job only.

## The change

`test-chapters` restores `~/.cache/huggingface` before the run and mounts it into the container as `HF_HOME`, so a checkpoint is downloaded once per key rather than once per job. `docker-compose.yml` already maps the same directory for a local run; CI was the one place that did not.

**Why a weekly key.** Nothing in the repo declares which checkpoints a chapter pulls: the model ids are literals inside the notebooks, and a manifest listing them would be a second place to keep in step. The key buckets by ISO week with a restore-key falling back to the previous week, so a newly added model is cached within a week at the cost of one download. A constant key would never save again after its first run and would freeze the cache at whatever the chapters needed that day.

**Why a coarse key is safe.** HuggingFace stores each snapshot under its own revision sha. A hit either holds the revision the notebook asks for or falls through to the network, so the key decides how much of the download is skipped and never which bytes load.

## Verification

`.github/workflows/test.yml` is in the `shared` path filter, so this PR runs the full chapter and case-study matrix. The first run populates every key cold and proves the mount and `HF_HOME` plumbing; the caching itself shows up as a hit on the run after it.

## Not in this change

An `HF_TOKEN` would raise the anonymous rate limit and is the belt to this PR's braces, but it needs a repository secret that does not exist yet.

https://claude.ai/code/session_01UPDj1nioTuD4xATMVyhZxB

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1489,6 +1489,22 @@ jobs:
               -v --tb=short \
               -k "$FILTER"
 
+      # The container is root and the runner is not, so whether the cache can be saved depends
+      # on the mode each writer happens to use. `huggingface_hub` writes 644 and tars fine; the
+      # `evaluate` library writes its downloaded metric scripts 600, and three of those under
+      # `evaluate/downloads/` failed ch10's save with `Cannot open: Permission denied` and tar
+      # exit 2 - reported as a warning, so the job stayed green having cached nothing. One
+      # chmod as root covers every writer instead of one per library. `a+rX` and not `a+rx`:
+      # the capital X sets the execute bit on directories only, leaving file modes alone.
+      #
+      # `always()` because a run that failed still downloaded the checkpoints, and they are the
+      # expensive part to fetch again.
+      - name: Let the runner read what the container cached
+        if: always() && steps.reach.outputs.have_key && steps.image.outputs.ref != ''
+        run: |
+          docker run --rm -v "$HOME/.cache/huggingface:/hf-cache" \
+            "${{ steps.image.outputs.ref }}" chmod -R a+rX /hf-cache
+
   # ---------------------------------------------------------------------------
   # Case study pipeline tests — inside Docker (ml4t/ml4t:latest)
   # ---------------------------------------------------------------------------
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1403,13 +1403,29 @@ jobs:
       # run from the previous week's entry instead of empty. A constant key would
       # never save again after the first run, freezing the cache at whatever the
       # chapters needed that day.
+      #
+      # Named jobs rather than all of them, and the list is a BUDGET decision with a number
+      # behind it. A repository gets 10 GB of Actions cache and this one was already holding
+      # 7.0 GB of buildkit layer blobs, the largest a single 4.4 GB base layer. Caching every
+      # chapter measured 3.83 GB and put the repository at 10.85 GB, over the limit, where
+      # GitHub evicts least-recently-used - which would have been those layers, silently
+      # trading a lockfile-image rebuild for a checkpoint download. The four that download
+      # anything at all are ch05, ch10, ch11-12 and ch13; ch10 alone is 2.66 GB of the 3.83,
+      # and it is the only one of the four never observed failing. Dropping it fits in 1.27 GB.
+      #
+      # Re-measure rather than trusting this list: `gh api .../actions/caches` shows what each
+      # key costs and `.../actions/cache/usage` shows the total. A chapter that starts pulling
+      # a checkpoint shows up as a key with a real size and belongs here; a chapter here that
+      # stops pulling one shows up as a ~200-byte key and does not.
       - name: HuggingFace cache key
         id: hfkey
         if: steps.reach.outputs.have_key
         run: echo "week=$(date -u +%GW%V)" >> "$GITHUB_OUTPUT"
 
       - name: Cache HuggingFace checkpoints
-        if: steps.reach.outputs.have_key
+        if: >-
+          steps.reach.outputs.have_key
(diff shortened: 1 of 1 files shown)


stefan-jansen:
## The mount and HF_HOME plumbing, verified locally against the real image

CI proves the whole path, but the two things that could silently do nothing - whether `HF_HOME` lands the cache on the bind mount, and whether the runner can then tar what the container wrote - are checkable here. Both hold.

```
$ docker run --rm -v "$D:/hf-cache" -e HF_HOME=/hf-cache ml4t/ml4t:latest \
    python -c "from huggingface_hub import hf_hub_download; \
               print(hf_hub_download('sentence-transformers/all-MiniLM-L6-v2','config.json'))"
/hf-cache/hub/models--sentence-transformers--all-MiniLM-L6-v2/snapshots/1110a243fdf4706b3f48f1d95db1a4f5529b4d41/config.json
```

The file is on the host under the mount, so `actions/cache` has something to save.

**The container runs as root** - `ml4t/ml4t:latest` declares no `USER`, and `os.getuid()` inside it is 0 - so everything it writes into the cache is root-owned on the runner. That is the one arrangement that could have made the post-job save fail silently, since `actions/cache` runs as the unprivileged runner user. It does not: root's umask leaves every directory 755 and every file 644, both readable by a non-owner, and `tar -cf /dev/null -C "$D" .` as uid 1000 over a root-owned tree exits 0.

Restore runs the other way and is also fine: `actions/cache` extracts as the runner user, and the container is root, so it can write over what it finds.

stefan-jansen:
## Correcting the comment above, and the table in the description

The first full run caught two things I had asserted rather than measured.

### `tar -cf /dev/null` cannot fail, so my verification proved nothing

I checked the save path with `tar -cf /dev/null -C "$D" .` and read exit 0 as "the runner can read what the container wrote". GNU tar special-cases `/dev/null` and never opens the members, so that command exits 0 over a tree it cannot read. Redone against a real archive, the same tree gives:

```
tar: ./f600: Cannot open: Permission denied
tar: Exiting with failure status due to previous errors
tar exit=2
```

**And that is what happened on ch10.** `huggingface_hub` writes 644 and tars fine, which is all my test exercised; the `evaluate` library writes its downloaded metric scripts 600, and three of those under `evaluate/downloads/` sank the save:

```
tar: ../../../.cache/huggingface/evaluate/downloads/5772...py: Cannot open: Permission denied
Failed to save: "/usr/bin/tar" failed with error: exit code 2
```

`actions/cache` reports that as a **warning**, so ch10 was green having cached nothing - the failure shape this change exists to prevent. `43bc74f2` adds one `chmod -R a+rX` as root after the run, which covers every writer rather than one per library, and runs on `always()` because a failed run still paid for the downloads. Verified against `ml4t/ml4t:latest`: exit 2 before, exit 0 after.

### Which chapters actually download, measured rather than grepped

The table in the description came from grepping `"org/name"` literals out of the notebooks. What the run actually cached:

| job | cached | |
|---|---|---|
| `ch11-12` | 777 MB | `finbert-tone`, as expected |
| `ch05` | 291 MB | **`distilgpt2`**, which the description misses entirely |
| `ch13` | 197 MB | `granite-timeseries-ttm-v1`, `chronos-t5-tiny` |
| `ch10` | save failed | the four text models and two datasets |
| `ch04`, `ch22`, `ch23` | ~200 bytes | download nothing in CI |
| every other `ch*` | ~200 bytes | - |

Two errors in one table. `ch05` was missed because `distilgpt2` is a bare model name with no org prefix and the regex wanted a slash. `ch04`, `ch22` and `ch23` name `all-MiniLM-L6-v2` in source but their notebooks are deselected or skipped in CI, so the grep found an id nothing fetches. Only four jobs of eighteen download anything, and the ones that do are the ones that were failing.

### Cache budget, since this is shared with every other cache the repo keeps

The four real entries total 1.24 GB. Repo-wide usage is now 8.86 GB against GitHub's 10 GB, across 91 caches. The weekly key means two generations of HuggingFace entries can coexist for one week's overlap; the older one is read at most once, by the first run of the new week through `restore-keys`, and then goes cold, so LRU eviction takes it before anything in active use. The fourteen ~200-byte entries cost an entry each and no meaningful space.

PR #1021, opened 2026-09-13
Chapter 20 cites sections that do not contain what is claimed

stefan-jansen:
Every section reference in the chapter-20 notebooks points at a section number that exists, so an existence check finds none of these. Only reading the target does. Eleven do not carry the claim.

**Option cost material sent to §20.5**, which is the allocator comparison and contains no cascade, no rungs and no bid-ask discussion. Six references, retargeted to §18.8 (the O'Donovan-Yu cost-mitigation cascade) or §20.6 (the cost-grid sweep).

**Wrong stage of the chapter.** The per-family IC comparison is Table 20.4 in §20.3, not §20.2 - and `ic_comparison.parquet` is read by notebooks 03 and 04, not by 02, so both halves of that reference were wrong. The validation-to-holdout decay analysis is §20.1, whose Figure 20.7 is the validation-versus-holdout scatter; §20.6 is trading realism.

**Wrong figure.** The figure paired with Table 20.6 in §20.5 is Figure 20.14. Figure 20.7 belongs to §20.1.

**A section that describes nothing of the kind.** §20.4 has no funnel - it covers the fundamental law, cadence and breadth, win rate, and family rankings. The pipeline funnel is Figure 20.18 in §20.9. The pointer is removed rather than retargeted: `sharpe_progression.parquet` is stage-by-stage Sharpe and Figure 20.18 is gate counts, so there is no section that describes it.

**Wrong chapter.** The PROCEED / REVISE / STOP triage was cited to Ch8 §8.5, which is cross-cutting feature types. The triage gates are §7.3 and the false-discovery control is §7.4; neither `PROCEED` nor Benjamini-Hochberg appears anywhere in chapter 8.

Two more surfaced while reading the targets:

- `08_recommendations` claimed §20.6-20.7. It traces the attrition funnel and classifies failures, which is §20.9 and its Figure 20.18. §20.6 and §20.7 already belong to notebooks 06 and 07, and §20.9 was cited by nothing.
- `07_regime_risk` gave §20.7 the title "Risk Overlays and Stability Across Regimes". The section is "Risk overlays" and carries no regime or stability material.

## A stale comment contradicting its own code

Two comments in `01_aggregate_synthesis` said the sp500_options headline is pinned to the Rung-2 full-universe baseline, with the Rung-3 liquid-subset variant reported elsewhere. The code does the opposite: `strategy_analysis.UNIVERSE_RESTRICTIONS` is `{"sp500_options": "liquid"}` and excludes full-universe rows from rank-1 selection, and §20.1's Table 20.2, §18.8 and §20.9 all carry the liquid-subset figure as the headline. The two numbers those comments quoted, -0.361 and +0.455, appear nowhere in chapter 20, so they are dropped rather than replaced with numbers this PR has not established as canonical.

## Cost

Markdown cells, code comments, and two docstrings in notebooks that carry no provenance stamp and no outputs. `notebook_provenance.py check` reports `0 stale, 0 outputs-changed` on all five, and the `.ipynb` diffs mirror the `.py` diffs line for line. No re-run.

    01_aggregate_synthesis   0 stale, 0 outputs-changed
    02_feature_evaluation    0 stale, 0 outputs-changed
    05_portfolio_allocation  0 stale, 0 outputs-changed
    07_regime_risk           0 stale, 0 outputs-changed
    08_recommendations       0 stale, 0 outputs-changed

`03_signal_quality` and `04_signal_to_strategy` are the two stamped notebooks in the chapter; every reference in them was checked against its target and is correct, so neither is touched.

https://claude.ai/code/session_01NReuqDy57ft7qkAAKUEtVG


stefan-jansen:
Measurement for whoever is tempted to turn this into a CI check, because the obvious check finds none of it.

**Every one of the 175 `§N.M` references in the notebook `.py` files points at a section file that exists.** Measured against `origin/main` today, resolving each against `book/<NN>_*/chapter/sections/section_<MM>_*.md`:

```
175 section references in notebook .py files
  0 pointing at a section number that does not exist
```

So every defect in this class is a reference that resolves to the wrong *true* section. Existence checking, the cheap instrument, cannot reach any of them. Only opening the target and reading it does - which is what this PR did.

Where they are, by chapter cited:

| ch | refs | | ch | refs | | ch | refs |
|---:|---:|---|---:|---:|---|---:|---:|
| 3 | 34 | | 2 | 33 | | 20 | 35 |
| 19 | 13 | | 6 | 9 | | 8 | 8 |
| 12 | 8 | | 25 | 8 | | 24 | 7 |
| 16 | 6 | | 18 | 5 | | 1 | 3 |
| 23 | 2 | | 5, 10, 15, 22 | 1 each | | | |

Chapter 20 carries 35 of the 175 and this PR finds 11 of them wrong. I would not extrapolate that rate - chapter 20 was audited because it was already suspected, so its rate is not an estimate for the other 140. But it does say the class is not rare enough to leave to incidental discovery, and that the instrument has to be reading rather than resolving.

Two of the eleven were also in cells I rewrote on #1015 and are already corrected there (`§20.2` to `§20.3` for the model-family IC pivot, which is Table 20.4; `§20.5` to `§20.6` for option execution costs, where §20.6's own text says the cost model must fit the instrument). If this PR rebases onto `c57b8eef` those two hunks will be gone rather than conflicting - worth reading rather than assuming.

Your reading of `01:2020`/`01:2021` is the better call than mine. I only checked that `-0.361` appears in no chapter-20 section file and left the comments alone; dropping them outright is right if they also invert `UNIVERSE_RESTRICTIONS`.

PR #1031, opened 2026-09-13
Fit every etfs latent model on the same folds, and see a fleet that does not

stefan-jansen:
`etfs/11c_conditional_autoencoder` and `etfs/11e_supervised_autoencoder` ran all eight seeded folds while the other nine `etfs` entries ran two. That is two costs, and only one of them is visible from the entry.

**The visible one.** On the `e2e649ed` run of `cs-etfs`, 11c took 501 s and 11e 521 s of a 3,041 s job. `cs-etfs` is the longest job in the matrix, so it sets how long `main` takes to report a full-matrix verdict. Fold cost rises with `fold_id`, so folds `[0, 1]` is below a quarter of that.

**The one that is not about cost.** `13_model_analysis` ranks these members against each other on their average per-fold IC, and its own prose states what differing fold sets do to that number: an average "taken over fewer folds than the others ... is not the same quantity". Two members at two folds beside a third at eight were being read as one comparison.

Neither is visible at the entry - both 11c and 11e look reduced, each carrying a `max_symbols` - so this also adds a check.

## The check

A case study's model entries must agree on one fold set. An entry that declares no `folds` counts as its own value rather than as an absence, because it runs every fold the fixture seeded, which is the state 11c and 11e were in.

Two case studies do not agree, and they are named in a ratchet with what disagrees rather than quietly excluded:

- `nasdaq100_microstructure` - `06_linear` and `07_gbm` declare `[0]`, the four sequence entries declare `[0, 1]`. Both linear entries carry a skip, so nothing compares them today and the disagreement costs nothing until that skip lifts.
- `us_firm_characteristics` - `05_linear` and `06_gbm` declare `[0]`, `07_tabular_dl` and `08a` through `08d` declare `[0, 1]`, and `10_model_analysis` ranks all of them together.

The test asserts the disagreeing set is exactly those two, so nothing joins it without a line and a case study that comes into line retires its own. DML entries are excluded because their resolver takes `n_folds`, a count, and refuses `folds` outright; a negative case pins that exclusion, alongside one for each of the two ways a fold set can disagree.

## Verification

`tests/test_pm_helpers.py`, `tests/test_overrides_schema.py`, `tests/test_smoke_configs.py` and `tests/test_model_registry.py`: 220 passed, 98 skipped. The ratchet was mutation-checked by reverting 11c's declaration in memory - it reports `etfs` and fails.

The reduced notebooks themselves are not run here. `cs-etfs` prices them, and the 900 s cap stays until it does: tightening it against a predicted figure rather than a measured one is the mistake the sibling `11d` entry's comment records.

https://claude.ai/code/session_01EKVuCYKyzMyRG6SEm8HwqU


stefan-jansen:
Measured on this PR's own `cs-etfs` job, against the `e2e649ed` run the PR was written from.

**`cs-etfs` went from 3,041 s to 1,580 s** (50:40 to 26:20). Not all of that is this change, and the log says how much is not: across the twenty unchanged notebooks in the same job the total fell from 1,974 s to 1,401 s, a **29 % faster runner**.

Adjusting for that:

| notebook | before | after | at the same runner speed | fold cut accounts for |
|---|---|---|---|---|
| `11c_conditional_autoencoder` | 501 s | **69 s** | ~356 s | 19 % |
| `11e_supervised_autoencoder` | 521 s | **74 s** | ~370 s | 20 % |

So the reduction is worth about 880 s of the 1,461 s the job dropped, and the rest was a quieter machine.

The prediction in the entry comment was 23 %, taken from the fixture's fold manifest - folds 0 and 1 hold 247,370 of 1,060,808 training rows. Measured 19 % and 20 %. Close enough that the manifest is a usable way to price a fold reduction before running one, which is the part worth keeping.

Both notebooks pass. The 900 s caps stay, as the entry says, until there is a figure to tighten them against - and now there is one, so tightening them is a follow-up with a measurement behind it rather than a guess.

PR #1040, opened 2026-09-13
Dataset cards: name a script that exists, read the keys the writer emits

stefan-jansen:
## What a reader hit

Every `data/**/dataset_card.py` ends with a profile section that either prints the dataset's profile or tells the reader how to get one. It was broken in both arms.

**The instruction was unrunnable.** Eight of the nine cards said `Generate with: python generate_profiles.py --dataset <name>`. No such script exists in this repository and none ever has. It rendered only in the `else` arm, so it was shown precisely to the reader who had no profile and wanted one.

**The other arm raised.** The cards read `profile['rows']`, `profile['columns']` and `profile['memory_mb']`, while `save_dataset_profile` writes `DatasetProfile.to_dict()`, whose keys are `total_rows`, `total_columns`, `columns`, `generated_at`, `source`, `date_range_start`, `date_range_end` and `symbols`. A card that did find a profile raised `KeyError` on its first printed line. The microstructure card was worse after the path was corrected: `profile.get("rows", "N/A")` missed, and `f"{'N/A':,}"` raises `ValueError: Cannot specify ',' with 's'`. The `factors` card failed silently instead, reading `profile.get("files", [])` and printing `Files: 0` for a profile of any size.

## What this does

Three helpers in `utils/downloading.py` so a reader and the writer cannot drift apart again:

- `dataset_profile_path` delegates to `get_profile_path`, the same rule `save_dataset_profile` writes by
- `load_dataset_profile` returns `None` rather than raising when there is no profile
- `print_dataset_profile` renders the keys `to_dict` actually emits

Every card now resolves its profile through those and, when there is none, either credits the script that writes it or says plainly that this dataset's downloader writes none and points at `save_dataset_profile`.

## The check

`tests/test_dataset_cards_name_real_scripts.py` asserts every `python <script>.py` a card tells a reader to type resolves against the repository root or the card's own directory, taking commands from string literals and fenced code blocks, and running nothing. Nothing else catches this: the string is in an `else` arm and a card is not executed in CI.

A second assertion refuses a card that both credits a writer and denies one, which is a contradiction a reader cannot resolve. It deliberately does not assert the underlying claim. Whether a script writes a profile is not decidable from its own text: `data/etfs/market/download.py` never names `save_dataset_profile` and writes one anyway, through `ETFDataManager.save`, which calls `generate_profile` and `save_profile`. An earlier revision of this branch used a grep for that symbol, got the etfs card wrong, and was caught by review; the reasoning is recorded in the test so the next reader does not re-derive it.

Both assertions have a companion that fails if the corpus or either form disappears, so neither can pass vacuously. Each was falsified by reintroducing the defect and confirming the named failure.

## Verification

20 passed, `ruff` clean, notebook pairs synced. No card anywhere still names `generate_profiles.py` or reads `rows`, `columns` or `memory_mb`.

## Provenance

This work sat uncommitted in a worktree from 2026-09-11, survived the 2026-09-13 machine crash, and is committed here unchanged apart from the microstructure and factors fixes and the checks above.


stefan-jansen:
Superseded by #1042.

#994 landed on `main` after this branch opened and solved seven of the nine cards with `ml4t.data.storage.data_profile.load_profile`, which is a better shape than the `utils/downloading.py` wrappers here. This branch is 54 commits behind with 14 conflicting files, and most of its diff is now a second answer to a solved question.

#1042 rebuilds the remainder on current `main`: the two cards #994 did not reach (`data/factors`, `data/equities/market/microstructure`, both still looking for a bare `profile.json` nothing writes, the second also carrying a latent `ValueError`), the two summary-table paths that contradict their own code (`data/etfs/market`, `data/fx/market`), and a test that makes all three decidable from the working tree.

Closing rather than resolving the conflicts. The branch stays on origin.

PR #1050, opened 2026-09-14
nasdaq100_microstructure: the deep-learning notebooks, the re-swept baseline stage, and the chapter analysis

stefan-jansen:
The `nasdaq100_microstructure` case study from the deep-learning notebooks through the backtest sweep and the chapter analysis. Opened while the last stages are still running so the required checks execute against a diff that is still small; 15 through 20 will land on this branch as they finish.

## What is here

**Engine lock.** Merged `origin/main` to take `ml4t-backtest` 0.1.6. The three fields it adds to `backtest_config` are defaults naming behaviour 0.1.3 already had, but they sit inside the object `backtest_hash` covers, so every freshly built spec hashed to an identity the registry did not hold. Resolved by re-keying the registry rather than re-sweeping it: verified on all 3,762 rows that a freshly built 0.1.6 spec hits the stored hash, and on a replay spanning both signal methods and all five families that the metrics are identical to the digit.

**Two-pass sweep.** The baseline pass runs every admissible prediction over the three equal-weight arms on the cost-feasible universe; the mechanism pass takes the top eight by baseline Sharpe over the full 45-arm grid. 846 cells, 0 failures.

**The ensemble producer.** `case_studies/utils/ensemble.py` builds the mean-forecast ensemble the case study features, which three notebooks read and nothing produced. It draws its members from the pool the sweep actually ran, not from the catalog filter, and excludes degenerate prediction sets.

**Selection is universe-scoped.** `resolve_best_predictions` was unscoped while the registry carries two universes, so a ranking could mix them. `queries.py` now supports asking for the unrestricted universe explicitly, with a test.

**A GBM booster-layout fix.** The trainer writes boosters to a path the loader did not check, so feature importance silently fell back.

**Analysis corrections across 13, 15, 16, 17 and 20.** Claims that contradicted the resolved carrier, counts that named a different population than the one computed, a capacity table that counted configurations while pooling statistics over checkpoints, and rolling-metric prose that inferred regimes from sign crossings. Each is described in its own commit.

## Notes for review

- The registry is not in this diff. It is a gitignored artifact tree; what is reviewable here is the source that produced it and the renders that report it.
- The renders are large and mostly mechanical. The source `.py` half of each pair is the readable diff.
- A "constant forecast" recurs in these commits as a specific thing: a prediction set holding one value per fold, which ranks nothing, fills a slot book once, pays almost no cost, and therefore tops a cost-dominated ranking. Several corrections here are about keeping those out of comparisons.

https://claude.ai/code/session_013itqoM3MqdVndYogncTe7U


stefan-jansen:
The degeneracy widening in `notebook_contracts._DEGENERATE_SUBQUERY` runs for every case study, not only this one, so it is measured against all nine registries before merging. It adds `abs(ic) < 1e-12 AND ic_std < 1e-12` alongside the existing `ic IS NULL`.

| case study | fold rows | excluded before | after | newly caught | smallest non-zero \|ic\| |
|---|---:|---:|---:|---:|---|
| nasdaq100_microstructure | 1,569 | 0 | 12 | **12** | 2.0048e-16 |
| us_equities_panel | 11,760 | 15 | 15 | 0 | 1.6614e-06 |
| us_firm_characteristics | 5,691 | 0 | 0 | 0 | 5.9011e-06 |
| sp500_options | 645 | 0 | 0 | 0 | 4.0948e-06 |
| cme_futures | 2,881 | 6 | 6 | 0 | 1.7868e-05 |
| etfs | 5,266 | 12 | 12 | 0 | 1.7500e-05 |
| fx_pairs | 7,730 | 4 | 4 | 0 | 2.3300e-05 |
| crypto_perps_funding | 1,555 | 2 | 2 | 0 | 6.5700e-06 |
| sp500_equity_option_analytics | 2,175 | 0 | 0 | 0 | 3.1941e-07 |

The four rows in the lower half were measured independently on the same registries and agree.

Every newly caught row is in this case study, which is the one that stores a constant fold's coefficient as machine epsilon rather than NULL. In the other eight the smallest non-zero coefficient anywhere is 3.1941e-07, five orders of magnitude above the threshold, so the band the new clause opens is empty on those registries rather than merely unexercised. That is the check the widening needed: it excludes what it was written to exclude and nothing else.

The other global change on this branch, `registry/queries._universe_filter_clause`, moves from an exact `json_extract(...) = ?` to a `COALESCE(..., 'full')` comparison. It fires only when a `universe_filter` is passed, and `UNIVERSE_RESTRICTIONS` is keyed on this case study alone, so `_universe_filter_clause(None)` returns the same empty clause the old code did for every other caller.

PR #1052, opened 2026-09-14
An interrupted sequence fit refits the folds it is missing, not all of them

stefan-jansen:
## The problem

A sequence run keeps the checkpoints it had written when it dies, but nothing read them. The reconstruction path validates the full declared fold list and refuses a tree holding fourteen of sixteen, and the fit path asked `run_dl_cv` for every fold. So a run killed in fold 15 of 16 still paid for all sixteen: on `us_equities_panel/09_dl_nlinear`, 23-26 hours to recover one fold.

## What the fit branch does now

1. **Adopts a dead attempt's staging tree**, or opens a fresh one.
2. **Asks which folds it already holds** in full.
3. **Clears the partial fold** it is about to refit.
4. **Refits the rest**, into the same tree.
5. **Promotes**, and republishes the whole declared population from the promoted tree.

### Which folds are reusable is decided by the code the promote uses

`_deep_checkpoint_defect` and `_darts_checkpoint_defect` now hold the per-checkpoint check that `validate_*_checkpoint_population` was doing inline. The validator raises on the first defect; the new `complete_*_checkpoint_folds` reports instead. A fold counts only when every declared checkpoint for it passes, digests and architecture included, so **a resume can never adopt a fold the promote would then reject**.

### Adoption is claimed with an flock

Two runs of one training hash can be in flight at once - `nb-run.sh` holds slots, not per-hash locks - and writing both into one tree promotes a `models/` that is corrupt and validates. The kernel drops an `flock` when its holder dies, however it died, so a tree whose lock is free has no live writer. A stale lock file proves nothing either way; only the lock decides.

The lock file sits beside the tree, not inside it, so the promote does not carry one into `models/`. It is removed **after** the promote and never in the `finally`: removing it while the tree is still adoptable would let a later run create a fresh file at the same path, flock that instead of the one the live run holds, and adopt a tree another run is writing.

### A partial fold is cleared before it is refit

Checkpoints are immutable. `write_deep_checkpoint` raises `FileExistsError` rather than overwrite one whose content differs, and a refit's weights differ from the interrupted attempt's because the resumed run reaches that fold from a different RNG state. Without the clear the resume dies on the first fold it tries to recover, after paying for the others.

### The two backends need opposite answers about the fold population

This is where three of the review rounds went, and it is worth stating plainly because the two look symmetric and are not.

| | native (PyTorch) | Darts |
|---|---|---|
| checkpoint metrics come from | prediction shards on disk under `save_dir/_incremental` | `cfg_slices`, in memory |
| are the adopted folds in them? | **yes** - `clear_fold_predictions` drops one `(config, fold)` glob at that fold's start, so only a refit fold loses its own | **no** - this run predicted nothing for them |
| coverage comparison expects | adopted + refit | refit only |
| tree validation expects | adopted + refit | adopted + refit |

Get the native side wrong and every checkpoint falls short of the expected population, is skipped, and `assemble_cv_result` raises with nothing to assemble. Get either tree validation wrong and the adopted folds are reported as undeclared artifacts on a tree that is correct - after the resume has paid to refit the rest.

The native path also **checks the shards are there before adopting a fold**. A fold whose shards have been pruned cannot contribute a checkpoint metric, so it is refit instead: one fold rather than the failure of the whole resume.

### The fold filter moved above the backend dispatch

`run_dl_cv` dispatches to `run_darts_cv` and returns. With the filter below that dispatch a Darts config was handed every split, so a Darts resume refit everything it had adopted and hit the immutable-checkpoint conflict on the first one.

## Verification

**Against real fitted state, not only fixtures.** The orphaned tree from an interrupted `patchtst` run (`dl-fixture-out/.../training/cec6d584eeac/.models.a4409611....tmp`) holds fold 0 complete at epochs 1 and 2 and fold 1 at epoch 1 only:

- `complete_deep_checkpoint_folds` names fold 0 alone
- it names nothing when the architecture is pinned to something else
- `validate_deep_checkpoint_population` still refuses the tree as a whole, which is the refusal that used to force the full refit

**25 tests**, and every substantive one confirmed to fail against a mutation of the thing it pins: dropping `already_fitted_folds`, dropping the partial-fold clear, skipping the sidecar digest, moving the lock removal into the `finally`, seeding the native expectation empty, adding the adopted folds to the Darts coverage comparison, and dropping them from either tree validation.

**What is not covered here:** a full kill-and-resume of a real sequence notebook. The unit tests pin the decisions and the wiring, and the report is verified against a real partial tree, but the first production resume will be the first end-to-end exercise of `run_dl_cv` writing missing folds into an adopted tree. `tests/test_darts_state.py`, `test_deep_learning_adapter.py`, `test_deep_learning_state.py`, `test_deep_model_state.py` and `test_sequence_identity_roundtrip.py` all pass unchanged (91 tests), so the uninterrupted path is not disturbed.

https://claude.ai/code/session_01D4vnGis1YXpAEWbvWVPx6N


stefan-jansen:
A production measurement of what this recovers, from the crash that motivated keeping the tree in the first place.

The 2026-09-13 20:36Z hard crash killed `us_equities_panel/10_dl_lstm` 2h35m in, at 7 of 16 folds, with no clean shutdown in the journal. Its staging tree held 139 checkpoints, 39 MB, last write 90 seconds before the machine died - the numbers already quoted in the `except Exception: raise` comment on `main`.

The relaunch ran 2026-09-13T21:59:33Z to 2026-09-14T07:35:24Z, 9.6 hours, and refit all 16 folds. Its own launch script records why: "deep_learning writes no candidate_fold_completions, so the 2.55 h of fit the crash took is paid again in full; nothing resumes." (Measured and written by the session that owns that case study, not by me.)

`10_dl_lstm` reaches the fit through `run_model_population` at `case_studies/us_equities_panel/10_dl_lstm.py:298`, so it is on the staged route this branch changes, and the crash left exactly the state `_claim_staging_tree` adopts: a `.models.*.tmp` tree whose writer is dead, so the `flock` is free. With this branch the relaunch would have adopted those 7 folds and refit 9.

Two things this does not claim. Walk-forward fold cost rises with fold_id, so 7 of 16 folds is well under 7/16 of the wall clock - the 2.55 h the peer measured is the number, not a proportional share of 9.6 h. And this is a reconstruction of what would have happened, not an observed resume: the production kill-and-resume named in the PR body is still outstanding.

PR #1057, opened 2026-09-14
A constant prediction whose IC stored as machine epsilon passed the degeneracy test

stefan-jansen:
`degenerate_prediction_sql` excludes a prediction set whose fold produced a constant prediction, by testing `fold_metrics.ic IS NULL`. That is how one code path records a correlation over a constant vector. Another path reduces the correlation numerically rather than short-circuiting on zero variance, and stores machine epsilon instead, so a test for NULL alone reads that prediction set as a valid model.

## What that let through

Measured across all nine registries on 2026-09-14.

`nasdaq100_microstructure` holds **no NULL fold IC at all**, and twelve prediction sets - LASSO and ElasticNet at alpha 0.01 and 0.1, on `fwd_ret_5m`, `fwd_ret_15m` and `fwd_ret_60m` - store a fold IC of **2.0e-16** with an IC standard deviation of 2.0e-17, over 22,286 decision times of 78,812. All twelve passed the test, joined the published `nasdaq100_microstructure-linear-validation-v1` population, and carry **396 backtests**.

A constant prediction leaves top-k selection with no ranking to apply, so the basket falls back to row order. The four configurations return identical Sharpes per label, which is the tell that one constant prediction was backtested four times:

| label | best Sharpe from a constant prediction | validation backtests that beat it | total on the label |
|---|---|---|---|
| `fwd_ret_60m` | 1.467 | 3 | 1,015 |
| `fwd_ret_5m` | 1.424 | 2 | 871 |
| `fwd_ret_15m` | 1.461 | 23 | 1,436 |

## The change

The test becomes NULL **or** an IC indistinguishable from zero, at 1e-12: six orders of magnitude above the epsilon values it catches, and five below the smallest real fold IC in the fleet - `latent_factors/cae` on `sp500_equity_option_analytics`, -3.2e-07 over 525 entities, which stays selectable and is the negative control in the test.

Outside nasdaq nothing moves. Fleet-wide exclusion goes from 39 prediction sets to 51; the twelve added are all nasdaq's.

| case study | prediction sets excluded | backtests they carry |
|---|---|---|
| nasdaq100_microstructure | 12 (was 0) | 396 |
| us_equities_panel | 15 | 2 |
| etfs | 12 | 0 |
| cme_futures | 6 | 48 |
| fx_pairs | 4 | 8 |
| crypto_perps_funding | 2 | 6 |
| sp500_equity_option_analytics, sp500_options, us_firm_characteristics | 0 | 0 |

## What it does not do

No result is recomputed, no fit is repeated, no row is deleted. The 396 backtests stand in the registry; they stop being selectable and stop appearing in any leaderboard built on the shared helper. No published carrier moves: nasdaq's is `deep_learning/nlinear` at 2.30 on the primary label `fwd_ret_15m`, well clear of 1.461.

## Verification

`tests/test_degenerate_prediction_epsilon.py`, three cases including the negative control. Confirmed it fails against the old rule, which returns only `null-fold` where the new one returns the two epsilon sets as well. `tests/test_backtest_coverage_exclusion.py` and `tests/test_incompletely_registered_predictions.py` pass unchanged, 20 tests total.


stefan-jansen:
Superseded by #1050, which carries the same fix from 2026-09-12 (803e765f, 6f4d8654) with a stronger predicate: `ic IS NULL OR (abs(ic) < 1e-12 AND ic_std < 1e-12)` rather than the mean-only test here. The conjunction is the right one - a fold whose per-time ICs are real but cancel to near zero is a bad model, not a degenerate one, and it should stay selectable so a leaderboard can show it losing. I did not check open PRs before building this.

The measurements in the description stand and are not duplicated there, so recording them here rather than losing them: fleet-wide the widening moves only nasdaq, 39 prediction sets excluded to 51; nasdaq's twelve carry 396 backtests; on fwd_ret_60m only 3 of 1,015 validation backtests beat the best of them; the nearest real fold IC in the fleet is latent_factors/cae on sp500_equity_option_analytics at -3.2e-07 over 525 entities.

PR #1060, opened 2026-09-14
A top-ten table showed five configurations, because a re-keyed identity ranks twice

stefan-jansen:
A configuration keeps its results when a spec field is added that it never set: the backtest identity is new, the numbers are not. `best` ranks over every generation in the registry, so both rows compete for the same ten places and a reader sees the same configuration repeated with nothing on the page saying why.

## What is there

Measured 2026-09-14 across the nine registries, keying on prediction and on the strategy view a reader sees. Every pair agrees on Sharpe.

| case study | stage | configurations | held twice |
|---|---|---|---|
| us_firm_characteristics | signal | 2,276 | 912 |
| us_firm_characteristics | allocation | 240 | 80 |
| etfs | signal | 1,767 | 424 |
| crypto_perps_funding | signal | 2,109 | 273 |

1,689 pairs. I diffed one: the two specs differ in `account.lock_notional_update_mode` unset against `position_legs`, and `position_sizing.share_rounding` unset against `nearest` - both previously implicit defaults that the config schema made explicit. A schema change re-keyed the identity and moved nothing measurable.

## What a reader saw

On the registries as they stand, before and after:

"Distinct" has two readings here and they give different numbers, so both are stated. **Rows a reader can tell apart** counts rows whose displayed columns differ. **Distinct configurations** counts distinct `(source, config_name, top_k, prediction_hash)`.

| case study | stage | label | rows a reader can tell apart | distinct configurations |
|---|---|---|---|---|
| us_firm_characteristics | allocation | fwd_ret_1m | 5 -> 10 | 5 -> 8 |
| us_firm_characteristics | signal | fwd_ret_1m | 5 -> 10 | 5 -> 10 |
| etfs | signal | all | 7 -> 10 | 7 -> 10 |

The allocation stage is the one place the two readings part, and the residual is the allocator gap recorded below. Verified against the live registry: the top ten holds ten rows over eight configurations, because `gbm/leaves_7_mse` at `top_k` 50 on prediction `81864e850fb5` appears at Sharpe 2.735138 and 2.405705, and `gbm/leaves_63_mse` at `top_k` 50 on `1aa0c28c324d` at 2.869934 and 2.326153. Each pair is two allocators whose metrics genuinely differ, so the rule correctly keeps both, and nothing on the page says which allocator produced which row.

`crypto_perps_funding` and the default-label tables were already ten of ten; its duplicates sit below the cut.

Only one label per case study moves at all; every other label returns byte-identical rows in the same order.

| case study | stage | label | rows before -> after |
|---|---|---|---|
| us_firm_characteristics | signal | fwd_ret_1m | 1802 -> 912 |
| us_firm_characteristics | allocation | fwd_ret_1m | 148 -> 93 |
| etfs | signal | fwd_ret_21d | 1812 -> 1026 |

148 goes to 93 rather than 80: thirteen pairs that look like duplicates differ somewhere a reader sees, and the rule keeps them.

## The rule

Drop a row only when every column the caller receives, except `backtest_hash`, equals one already kept. Such a row carries nothing a reader could have read off it, so nothing is lost, and it is the conservative direction: four of the measured pairs agree only past the sixth decimal and are **not** collapsed, because the rule keys on the value rather than on how it renders.

The first row of any group survives under the query's existing `sharpe DESC, backtest_hash ASC` ordering, so nothing that selects on `best` can change its answer. Measured old against new on the three affected registries: the top row and its backtest hash are identical in all six stage tables.

This is the same defect `signal_method` alone had (#910), one level down. There the displayed columns could not separate rows that genuinely differed; here the rows do not differ at all.

The SQL `LIMIT` moves to the frame, because deduplicating after a LIMIT of ten returns fewer than ten rows. The full stage is a few thousand rows; the queries measure 0.05 to 0.58 s.

## Verification

`tests/test_leaderboard_identical_rows.py`, five cases: the collapse, a negative control across four columns that must each keep both rows, the sub-display-precision float pair that must stay two rows, nulls, and order preservation. Every existing test that touches `BacktestExplorer` passes - 113 tests across 17 files.

## Not fixed here

`best` returns `gbm/leaves_7_mse` at `top_k` 50 six times at six different Sharpes on `us_firm_characteristics`. Those are six checkpoints of one configuration and the checkpoint is not a displayed column, so nothing in the frame says what separates them.

That is the uncapped shape, not what ships. `11_backtest` diagnosed it at its line 495 and caps the table at `MAX_SLOTS_PER_CONFIG = 2`, printing the cap underneath - "At most 2 slots per configuration: 6 configurations, 1 family" - so a reader of the rendered page is told what they are looking at. Before that cap, `gbm/leaves_7_mse` held eight of ten slots and two configurations held all ten.

Two things remain true anyway: the cap lives in that one notebook rather than in the shared helper, so the other signal tables do not have it; and neither the cap nor the caption says which dimension separates two slots of one configuration. Adding `checkpoint_kind` and `checkpoint_value` to `best` answers both, and it means teaching more than twenty test fixtures that `prediction_sets` carries those columns - eleven tests across five files fail without it. That is its own piece of work with its own review, not something to attach to a deduplication.


stefan-jansen:
## The allocation stage carries six allocators per prediction, and none of them is a displayed column

Raised in review: the returned columns carry no allocator, so the collapse relies on two allocators never landing on bit-identical metrics for the same source, label and `top_k`. Measured rather than argued.

**The exposure is total, not rare.** Grouping allocation-stage rows by `(prediction_hash, allocation.top_k)`:

| case study | allocation rows | groups | groups holding more than one allocator |
|---|---|---|---|
| us_firm_characteristics | 320 | 120 | 120 |
| etfs | 378 | 63 | 63 |
| crypto_perps_funding | 720 | 40 | 40 |

Every group. `etfs` and `crypto_perps_funding` carry all six - `conformal_weighted`, `hrp`, `inverse_vol`, `mvo_ledoit_wolf`, `risk_parity`, `score_weighted` - on every prediction; `us_firm_characteristics` carries `conformal_weighted` and `score_weighted`. `strategy.allocation.method` lives inside `spec_json` and is not selected, so a reader of this table sees up to six rows for one prediction with nothing on the page saying what separates them.

**No pair of allocators produces identical metrics.** Joining `backtest_runs` to `backtest_metrics` and keying on `(prediction_hash, top_k, sharpe, cagr, max_drawdown, total_return, volatility)`:

| case study | metric-identical groups spanning more than one allocator |
|---|---|
| us_firm_characteristics | 0 |
| etfs | 0 |
| crypto_perps_funding | 0 |

Zero across 1,418 allocation rows. So the deduplication does not merge across allocators anywhere in the registries as they stand, and the five metrics agreeing to the bit while the portfolios differ is not something these strategies do.

**What this does and does not settle.** It settles that the change is safe on this data. It does not make the allocator a displayed column, so the collapse rests on a property of the values rather than on a key. That is the same shape as the checkpoint gap already recorded above under "Not fixed here": a dimension that separates rows is absent from the SELECT, and the fix is to add it rather than to rely on the metrics diverging. Adding `allocation.method` alongside `checkpoint_kind` and `checkpoint_value` is the one change that answers both, and it is the same piece of work with the same fixture cost.

PR #1086, opened 2026-09-15
The session output dir survives its own first teardown

stefan-jansen:
`seeded_output_dir` is session-scoped: it installs `ML4T_OUTPUT_DIR` exactly once, in whichever test first requests it, and `tests/conftest.py` restores that value after every test so the install survives.

Two test modules defined their own autouse fixture named `_restore_output_root`. An autouse fixture of that name in a test module **shadows the conftest one for every test in the module**, and both copies popped the variable unconditionally. So the very test that installed it removed it for the rest of the worker.

## Measured, not inferred

Sampling the variable at each phase across two tests, before the fix:

    setup-post  test_model_request_catalog_...   env='/tmp/pytest-of-stefan/pytest-427/ml4t_output0'
    td-post     test_model_request_catalog_...   env=None
    setup-pre   test_us_equities_pilot_...       env=None
    td-post     test_us_equities_pilot_...       env=None

and after:

    td-post     test_model_request_catalog_...   env='/tmp/pytest-of-stefan/pytest-428/ml4t_output0'
    setup-pre   test_us_equities_pilot_...       env='/tmp/pytest-of-stefan/pytest-428/ml4t_output0'

## Why it looked like missing artifacts

With the variable gone, every later test resolves through `get_case_study_dir` into the committed `case_studies/` tree instead of the seeded output dir. In a checkout whose `case_studies/*/` artifact symlinks exist that is invisible, because resolution finds real artifacts anyway. In one without them the loader raises:

    FileNotFoundError: Missing prerequisites for 'us_equities_panel':
    features/financial.parquet, labels/fwd_ret_1d.parquet

with both files present on disk. The same 15-file reproduction gives **320 passed** in a worktree with the links and **1 failed, 318 passed** in one without, which is why the reproduction did not travel between two sessions' checkouts.

## Three changes

Both module-level shadows are removed, so the conftest fixture applies where it was written to apply. Its own docstring says it lives there "because thirteen copies of it drifted apart once already"; two were still shadowing it.

The conftest fixture no longer restores `previous` blindly. It prefers the session value when one is installed, which is the rule the other conftest restorer already applied - otherwise the first test to request the fixture is exactly the test whose `previous` was `None`, so it pops what it just installed.

The guard is a **name** check rather than a behaviour check, deliberately: when a module shadows the fixture, the conftest version never runs, so there is no behaviour left in that module to observe. It fails with either shadow put back and names the file and line.

## Not this

Two mechanisms were ruled out by measurement before this one: a collection-time environment leak (every `ML4T_*` write under `tests/` is inside a function or class body, so none fires at import), and the files that sort before `test_artifact_specs.py` (running those sixteen in order passes). Both were paid for by others; recording them so nobody pays twice.

https://claude.ai/code/session_01H1MeiMToTiM2yAP34At4G9


stefan-jansen:
Closing in favour of #1087, which fixes the same defect and is the superset.

agents-10's objection to my `tests/conftest.py` hunk is correct and I measured it rather than conceding it. On a tree with both module shadows removed and `tests/conftest.py` exactly as it is on `main`:

    setup-pre  test_model_request_catalog_...  env=None   sess=None
    td-post    test_model_request_catalog_...  env='/tmp/pytest-of-stefan/pytest-444/ml4t_output0'
    setup-pre  test_us_equities_pilot_...      env='/tmp/pytest-of-stefan/pytest-444/ml4t_output0'
    td-post    test_us_equities_pilot_...      env='/tmp/pytest-of-stefan/pytest-444/ml4t_output0'

The variable survives its own first teardown with conftest untouched, so the conftest restorer is not a second poisoner and my trace measured the shadow, not it. The reason is fixture setup order: session-scoped fixtures are set up before function-scoped ones, so for the test that first requests `seeded_output_dir` the autouse fixture's `previous` is already the seeded dir. The `previous is None` branch is unreachable for that test, which makes my change an unreachable branch rather than hardening, and this PR body's claim that "the code produced it whenever the variable had been unset before the run" is wrong.

#1087 also covers a third file this PR leaves in place, `tests/test_label_digest_without_a_sidecar.py`, whose fixture pops the variable under its own name rather than shadowing - my guard keys on the conftest names and misses it.

The comment left at each de-shadowed site is worth keeping and agents-10 is taking it into #1087.

PR #1091, opened 2026-09-15
The provenance gate has not covered a merge in ten, and now fires on main

stefan-jansen:
The step `Notebooks this change touched were executed from their current .py` states its own purpose: *"a stale render reaches a reader through a merge, and no merge was covered"*, and *"a gate that only fires on the path we believe is the only path is how that happened."* It then fires only on `pull_request` and `push` - and on `main` that is the path it does not reach.

## Mechanism

Every commit on `main` gets two Tests runs: the `push` run, and the run `Docker Publish` triggers on finishing. They share one concurrency group with `cancel-in-progress`.

- The **`push`** run is the only one whose event carries a previous tip to diff, and it is routinely cancelled partway by the run that replaces it.
- The **`workflow_run`** run survives, and skipped this step, because `workflow_run` is neither name in the condition. A skipped step is a green job.

## Measurement

Across the last ten commits on `main`:

| | |
|---|---|
| `guards=[success, cancelled]` | **9** - the cancelled entry is this step's run; the success is the run that skipped it |
| `guards=[success, failure]` | **1** (`b6a7627c`) - the only commit where the step actually ran |

One in ten, and the one that ran caught a stale render immediately, which #1090 then fixed. Reading `guards=success` on a `main` commit has not meant the provenance gate passed.

## The change

On `workflow_run` there is no previous tip, so there is no honest narrower scope and it checks every notebook - the same fallback a first push already takes. The `pull_request` and `push` paths are untouched.

## Verification

Ran the exact command the new path runs, against current `main`: **exit 0**, `0 stale, 0 contradicted, 0 outputs-changed`. This adds coverage without turning `main` red.

https://claude.ai/code/session_01BR5oE2SZGDVY1y5qAPVT6o


stefan-jansen:
## Added: the checkout ref, which was a second half of the same defect

Raised in review against the first commit, and confirmed from the job log rather than reasoned about. On `workflow_run` the guards checkout runs `git rev-parse refs/remotes/origin/main` then `git checkout --force -B main refs/remotes/origin/main` - the branch tip at **fetch** time - while the check-run is stamped on `github.sha`, the tip at **event** time.

Those differ whenever `main` moves while the job waits on the concurrency group, and that window is not theoretical: the guards job on `b6a7627c` waited 35 minutes between its run being created at 07:32:53Z and the checkout at 08:08:56Z. The harmful direction is the stamped commit dirty and the fetched one clean - a green mark published against a tree nobody read.

**Three shas are in play on this event, not two**, which is where the first attempt went wrong. It pinned to `github.event.workflow_run.head_sha`; that is the upstream Docker Publish commit, a third sha again, so it would have fixed the checkout and left the same mismatch pointing the other way. The property that has to hold is that the tree read is the tree the result claims, so the pin is `github.sha`.

Empty is `actions/checkout`'s own default, so the `pull_request` and `push` paths resolve exactly as before.

Both expressions are single-line on purpose: a folded `>-` scalar keeps the newline on a more-indented continuation line instead of folding it, which puts a literal newline inside the `${{ }}`. Checked by parsing the file and asserting neither string contains one.

PR #1109, opened 2026-09-18
fx_pairs' allocation guard compared engine versions, not strategies

stefan-jansen:
`fx_pairs/14_portfolio_management` refuses an allocation whose non-allocation strategy fields
differ from its equal-weight baseline. The check is right and its input was wrong: it was
comparing which engine version serialized each row.

`ml4t-backtest` 0.1.3 to 0.1.6 added `account.lock_notional_update_mode` and
`position_sizing.share_rounding`, both previously implicit defaults that the config schema made
explicit. Every `fx_pairs` signal row was registered before 2026-09-12 and lacks both; every
allocation row computed after it carries both. So:

```
RuntimeError: allocation 510654bf3131 changed a non-allocation strategy field against
baseline 042f2cee8082:
  backtest_config.account.lock_notional_update_mode: baseline=None allocation='position_legs';
  backtest_config.position_sizing.share_rounding: baseline=None allocation='nearest'
```

## Measured

Against `fx_pairs`' registry, 2026-09-18:

| stage | rows | lacking both fields |
|---|---:|---:|
| signal | 1,932 | 1,932 |
| allocation | 1,111 | 229 |
| risk_overlay | 126 | 126 |
| cost_sensitivity | 110 | 77 |

Baseline `042f2cee8082` is a signal row from 2026-09-06T19:12:36Z. Allocation `510654bf3131` is
from 2026-09-15T20:51:42Z. The latest signal row in the case study is 2026-09-11, so there is no
baseline anywhere in the registry that carries the fields and no allocation run after 2026-09-12
can pass.

That the two fields moved nothing measurable is already recorded in the tree, from a different
measurement: `backtest_explorer._drop_rows_a_reader_cannot_tell_apart` documents the 2026-09-14
sweep of all nine registries, where 1,689 leaderboard pairs were identical, the single pair that
differed differed in exactly these two fields, and it agreed on Sharpe.

## The fix

`16_costs.py` and `19_strategy_analysis.py` hit this in 2026-09-12 and both round-trip the two
sides through the installed schema, which states the comparison in one vocabulary and covers the
next added field without naming it. `_non_allocation_projection` is the third site and was
missed. It now does the same thing, in the same shape, including merging the popped metadata back
over the serialized view.

`ensure_backtest_spec` deliberately does not round-trip, because its result is hashed and a
dropped unknown key would move an identity. Here the result is compared and discarded.

## Tests

Two added to `tests/test_fx_allocation_selection.py`, which already exercises this projection:

- a spec lacking the two keys projects equal to one carrying the schema defaults. **Fails on
  `main`**, with the diff naming exactly `lock_notional_update_mode` and `share_rounding`.
- a moved `account.allow_short_selling` still makes the projections differ, so forgiving an
  absent field did not make the account block unreadable. Passes either way; it is there so the
  first test cannot be satisfied by deleting the check.

`uv run pytest tests/test_fx_allocation_selection.py` - 9 passed.

## Provenance

The `.ipynb` is synced with `sync-inert`, not re-executed. The projection's only consumer is an
equality test; the edit applies the same deterministic normalization to both sides, so a run that
found two specs equal still does. The stamped 2026-09-11 run predates both fields, so both sides
lacked them and compared equal without the round-trip. Nothing the projection returns is printed.
`notebook_provenance.py check --since main` passes.

## How it was found

The private allocation-stage expansion in `ml4t/evidence` runs each case study's allocation
notebook over every advancing prediction set rather than the declared ten. `fx_pairs` was the one
of five that still refused on 2026-09-18, with this error, and it is the only one of the seven
allocation notebooks that carries this sibling check - the other six would not have noticed.


stefan-jansen:
Independent confirmation of the premise, measured today against other case studies' registries. (Corrected: an earlier version of this comment said 5,003 rows, which came from a count that over-reports. The measured figure is 4,982.)

Grouping `backtest_runs` on `(prediction_hash, artifact_digests_json)` alone does not isolate the effect. Identical artifacts on an identical prediction is not an identical spec: a risk overlay that never triggers reproduces the configuration without it byte for byte, and those are two genuine configurations rather than one re-keyed. A group counts here only when its members' specs differ **exclusively** in `backtest_config.account.lock_notional_update_mode`, `backtest_config.feed.vwap_col`, `backtest_config.position_sizing.share_rounding`, and the `metadata.preset_path` that `specs.py:203` strips before hashing.

Counted that way, across the nine canonical registries:

| registry | re-keyed groups | rows | identical output, different config |
|---|---:|---:|---:|
| sp500_equity_option_analytics | 2,491 | 4,982 | 7 |
| us_firm_characteristics | 911 | 1,822 | 0 |
| etfs | 789 | 1,578 | 0 |
| fx_pairs | 0 | 0 | 4 |
| the other five | 0 | 0 | - |

Every pair in those 4,191 groups carries byte-identical output parquets and the same metrics. An engine that gained defaulted config keys re-keyed three registries without changing a single number, so a difference in these fields is a schema difference and not a strategy difference. That is the premise this PR rests on.

The engine confirms the mechanism directly. In `ml4t-backtest` 0.1.6's `config.py`, the two fields are dataclass defaults (`:544` `POSITION_LEGS`, `:738` `NEAREST`); `to_dict` writes both unconditionally (`:920`, `:937`); `from_dict` reads them with those same defaults (`:1165`, `:1186`). An omitted key and an explicitly written value therefore build the same config object, so a guard comparing un-round-tripped specs compares spellings rather than configurations.

Note that `fx_pairs`' own registry is clean. That is consistent with, not contrary to, the failure this PR fixes: its baselines were registered before the engine wrote these keys, and today's allocation run computes with an engine that writes them, so the mismatch is between vintages rather than within the store.

What the table does not establish is the asymmetry this PR actually repairs, that `14_portfolio_management.py` compares an un-round-tripped config while `16_costs.py` and `19_strategy_analysis.py` normalize both sides through `EngineBacktestConfig.from_dict(...).to_dict()` first. That rests on a reading of those three files and on the 1,689 leaderboard pairs in `backtest_explorer._drop_rows_a_reader_cannot_tell_apart`, cited above.

